PR #1, opened 2020-07-02
MCS-263: Ability to add scene history files to Elasticsearch

brianpippin:
Ignore all the files in the folder Eval1, they are just files moving from the main MCS repo to this repo, and am doing this as more of an archival issue.


eval1/create_fake_ground_truth.py
```diff
@@ -0,0 +1,42 @@
...
+        ground_truth = "ground_truth.txt"
+        with open(ground_truth, 'w') as outfile:
+            for block in range(0, 3):
+                for test in range(0, 1080):
+                    results = [0, 0, 0, 0]
+                    indexes = random.sample({0, 1, 2, 3}, 2)
+                    results[indexes[0]] = 1
+                    results[indexes[1]] = 1
+                    for index in range(4):
+                        outfile.write("O{}/{:04d}/{} {}\n".format(block+1, test+1, index+1, results[index]))
```
chrislong, line 37: You might want to use format strings, instead. (Maybe not worth rewriting, but for future reference.)

mcs_scene_ingest.py
```diff
@@ -0,0 +1,57 @@
...
+    with open(os.path.join(folder, file_name)) as json_file:
+        return json.load(json_file)
+
+
+def ingest_files(folder: str, eval_name: str, performer: str) -> None:
+    schema = mcs_scene_schema.get_scene_schema()
+    #TODO: Update this file so that you dynamically can set an arg for mcs_scenes, scene_history, etc, that will use an appropriate scheme file
+    elastic_search = MCSElasticSearch("mcs_scenes", eval_name, True, schema)
+
+    scene_files = [f for f in os.listdir(folder) if "debug" in str(f) and str(f).endswith(".json")]
```
ThomasSchellenbergNextCentury, line 19: Right now they all end with `-debug.json` so that could make this simpler.

mcs_scene_schema.py
```diff
@@ -0,0 +1,206 @@
+from typing import Dict, Any
+
+_scene_schema = {
```
ThomasSchellenbergNextCentury, line 3: So Brian did a killer job at listing out all the properties of a scene here, but, do we need to keep all of them?  My initial impression is that we only really need the properties that we'll want to show in the UI.  I'm not sure what that is yet (maybe just the tags?), but it's probably not everything.  Plus, the more properties we want to keep, the more maintenance we need to do, since we'll want to add any new scene properties (that we add in the future) to the ES mapping as well.

On the other hand, there's a benefit to keeping everything here, and then cutting down if/when needed.  My only concern with keeping everything is that, if the ES records are bigger, the runtime performance may go down (though, with a smaller dataset like ours, it may not be noticable).

mcs_scene_schema.py
```diff
@@ -0,0 +1,206 @@
...
+                                },
+                                "scale": {
+                                    "properties": {
+                                        "x": {"type": "double"},
+                                        "y": {"type": "double"},
+                                        "z": {"type": "double"}
+                                    }
+                                }
+                            }
+                        },
```
ThomasSchellenbergNextCentury, line 118: Seems to be a redundant `shows` mapping here.

mcs_scene_schema.py
```diff
@@ -0,0 +1,206 @@
+from typing import Dict, Any
+
+_scene_schema = {
+    "mappings": {
```
ThomasSchellenbergNextCentury, line 4: I'm seeing a variety of smaller issues with the mapping (mainly some missing properties).  Rather than list each possible property like this, you could make use of ES's dynamic templates.  The built-in dynamic template works well for ints, doubles, and booleans, but sets all strings to `text`, which is not what you want.  You can override the built-in behavior by adding a `dynamic_templates` array directly inside the `mappings` object:

```json
    "mappings": {
      "dynamic_templates": [
        {
          "strings": {
            "match_mapping_type": "string",
            "mapping": {
              "type": "keyword"
            }
          }
        }
      ]
    }
```

mcs_scene_ingest.py
```diff
@@ -0,0 +1,57 @@
...
+        print("Ingest scene file: {}".format(file))
+        scene = load_scene_file(folder, file)
+        scene["eval"] = eval_name
+        scene["performer"] = performer
+
+        #TODO: update index_dict to use args depending on what you are ingesting so this file is more dynamic
+        index_dict = {
+            "index": {
+                "_index": "mcs_scenes",
+                "_type": "scenes"
```
ThomasSchellenbergNextCentury, line 34: The use of a `_type` here indicates to me that we're using ES 6 rather than ES 7.  There's nothing inherently wrong with ES 6, but we may want to consider upgrading to the latest version of ES.  I don't see anything here that would be incompatible with ES 7.

mcs_scene_schema.py
```diff
@@ -0,0 +1,206 @@
...
+                        "action_list": {"type": "nested"},
+                        "info_list": {"type": "keyword"},
+                        "series_id": {"type": "keyword"}
+                    }
+                },
+                "answer": {
+                    "properties": {
+                        "choice": {"type": "keyword"}
+                    }
+                },
```
ThomasSchellenbergNextCentury, line 198: The `"answer"` for interactive scenes will include an optimal path, rather than a `"choice"`.

mcs_scene_schema.py
```diff
@@ -0,0 +1,206 @@
...
+                        "category": {"type": "keyword"},
+                        "domain_list": {"type": "keyword"},
+                        "type_list": {"type": "keyword"},
+                        "task_list": {"type": "keyword"},
+                        "description": {"type": "keyword"},
+                        "metadata": {
+                            "properties": {
+                                "objects": {"type": "keyword"}
+                            }
+                        },
```
ThomasSchellenbergNextCentury, line 187: The `"metadata"` for interactive scenes will include one or more target(s), and they're not called `"objects"`.  Additionally, each target will have its `"image"` pixel array, which we'll probably want to remove so we don't bloat the ES index.

mcs_scene_ingest.py
```diff
@@ -0,0 +1,57 @@
...
+    scene_files = [f for f in os.listdir(folder) if "debug" in str(f) and str(f).endswith(".json")]
+    scene_files.sort()
+
+    ingest_scenes = []
+
+    for file in scene_files:
+        print("Ingest scene file: {}".format(file))
+        scene = load_scene_file(folder, file)
+        scene["eval"] = eval_name
+        scene["performer"] = performer
```
ThomasSchellenbergNextCentury, line 28: In a future ticket, we'll want to add links for any media we save in S3 that are relevant to this scene (images of target objects, videos of the optimal path, etc.)

mcs_scene_schema.py
brianpippin: My plan at the moment was to keep all the mappings for now, and then once we get everything in for the psychologists to look at possibly going back and modifying the schema in which case we would then have to be more diligent about creating the object being ingested and not just ingest the whole json file.

I am planning to ingest all the training data and the test data that Dean generated this weekend, so I plan to modify the schema with anything I missed.

mcs_scene_ingest.py
brianpippin: That is true, I suppose it would make sense to just change it to that.

mcs_scene_ingest.py
brianpippin: We have a ticket to upgrade our Neon version, which I assume the new version of Neon is using ES7?  I don't know if this something we want to tackle at this point but I assume if we can find some time to carve out it would make sense to upgrade to the latest.

mcs_scene_schema.py
brianpippin: Yes working on adding in interactive scene properties this weekend.

mcs_scene_schema.py
brianpippin: We definitely do not want the pixel array in ES, that would severely bloat our ES

mcs_scene_ingest.py
brianpippin: We will either link it, or make sure we save the folders based on something like evalname-performer-scene in which case we could build the folder path with things we already have.  Either way would work.

brianpippin:
All, I would say you can start looking at this, I am not sure I am going to be working on it much more until I refactor.  The main file to look is probably mcs_scene_ingest.py.  It is nice having the schema files and that is helpful for when you need to update the config.yaml for Neon.  Also anything in the EVAL1 folder is LEGACY code from eval1, I didn't touch it, just moved it out of the MCS repo and wanted to have it stored.  If something ever clears the Elasticsearch cluster, it might be necessary to reingest eval1 data.

mcs_elasticsearch.py
```diff
@@ -0,0 +1,28 @@
...
+        self.es = Elasticsearch([{'host': MCSElasticSearch.elastic_host, 'port': MCSElasticSearch.elastic_port}])
+
+        # delete index if exists
+        if delete_index:
+            if self.es.indices.exists(index_name):
+                print("Removing existing index " + index_name)
+                self.es.indices.delete(index=index_name)
+
+        # create index
+        self.es.indices.create(index=index_name, body=settings, ignore=400,)
```
ThomasSchellenbergNextCentury, line 22: Extra comma here at the end

mcs_elasticsearch.py
```diff
@@ -0,0 +1,28 @@
+from elasticsearch import Elasticsearch
+from typing import Dict
+
+class MCSElasticSearch(object):
+    elastic_host = 'localhost'
+    elastic_port = 9200
+    index_name = ''
+
+    def __init__(self, index_name: str, index_type: str, delete_index: bool, settings: dict):
```
ThomasSchellenbergNextCentury, line 9: We're not using index_type here

mcs_scene_ingest.py
```diff
@@ -0,0 +1,340 @@
...
+    return history_files
+
+
+def get_index_dict(index: str, index_type: str) -> dict:
+    return {
+        "index": {
+            "_index": index,
+            "_type": index_type
+        }
+    }
```
ThomasSchellenbergNextCentury, line 77: FYI In ES 7+, indexes no longer have `_type` properties, and thus the data in our indexes can't have multiple different "types", so we shouldn't rely on that in our data schema (for example, to differentiate data from different evals).  We can either split them into entirely separate indexes and/or add differentiating properties (like an "eval" property).  We can probably handle this in a future ticket, though.

mcs_scene_ingest.py
```diff
@@ -0,0 +1,340 @@
...
+        history = load_history_file(folder, file)
+
+        history_item = {}
+        history_item["eval"] = eval_name
+        history_item["performer"] = performer
+        history_item["name"] = get_scene_name_from_history_file(file, "-202.+-")
+
+        history_item["test_type"] = history_item["name"][:-7]
+        history_item["scene_num"] = history_item["name"][-6:-2]
+        history_item["scene_part_num"] = history_item["name"][-1:]
```
ThomasSchellenbergNextCentury, line 124: Ah, I get it.  All of this info is available in the scene file, but not in the history file.  Should be easy enough to do for the next eval.  Sorry for the trouble!

ThomasSchellenbergNextCentury: I see that on line 158, you load the corresponding scene file.  You should be able to pull the `name` out of this file.

ThomasSchellenbergNextCentury: So the name is currently created using a custom prefix, the scene number, and the pair/quartet number.  I guess it would be helpful if we had separate properties for each of these strings?

brianpippin: Yeah I was just thinking having separate properties so we don't need to do any string manipulation at all.

mcs_scene_ingest.py
```diff
@@ -0,0 +1,340 @@
...
+                    else:
+                        history_item["category"] = "interactive"
+                        type_parts = history_item["test_type"].rsplit('-', 1)
+                        history_item["category_type"] = type_parts[1]
+                        history_item["category_pair"] = type_parts[0]
+                else:
+                    history_item["category"] = "interactive"
+                    type_parts = history_item["test_type"].rsplit('_', 1)
+                    history_item["category_type"] = type_parts[1]
+                    history_item["category_pair"] = type_parts[0]
```
ThomasSchellenbergNextCentury, line 173: `scene['goal']['category']` will either be `intphys`, `traversal`, `retrieval`, or `transferral`.  For the `intphys` scenes, `scene['goal']['type_list']` is a list that will have either `object permanence`, `shape constancy`, or `spatio temporal continuity` in it.

ThomasSchellenbergNextCentury: `scene['goal']['type_list']` will also have either `interactive` or `passive` in it.

mcs_scene_ingest.py
```diff
@@ -0,0 +1,340 @@
...
+                    if history_item["score"]["confidence"] == 1 and history_item["score"]["classification"] == "implausible":
+                        history_item["score"]["adjusted_confidence"] = 0
+                    elif history_item["score"]["classification"] == "implausible" and history_item["score"]["confidence"] > 0.5:
+                        history_item["score"]["adjusted_confidence"] = 1 - history_item["score"]["confidence"]
+                    elif history_item["score"]["confidence"] == 1:
+                        history_item["score"]["adjusted_confidence"] = 1
+                    else: 
+                        history_item["score"]["adjusted_confidence"] = float(history_item["score"]["confidence"])
+
+                    history_item["score"]["mse"] = math.pow((history_item["score"]["ground_truth"] - round(float("{0:.9f}".format(history_item["score"]["adjusted_confidence"])))), 2)
```
ThomasSchellenbergNextCentury, line 202: This is the scoring for the passive scenes.  We're still waiting to get the scoring algorithm from Amir for the interactive scenes.  I don't think that's going to happen at this point, though, at least for this eval.  My hope was that we could use each interactive scene's optimal path to calculate a percentage that we could use to weight the score.  Maybe it's too late to do that now.

mcs_scene_ingest.py
ThomasSchellenbergNextCentury: I'm not sure if it would be worth separating these tags into their own distinct properties, or if just using the `type_list` is good enough.

PR #3, opened 2020-11-30
MCS-443, 444, 445, 446 - Updating Ingest to mongo, removing Elastic search, updating for Eval 1 as well.

brianpippin:


rartiss55 (approved):

eval1/create_json_ingest.py
```diff
@@ -139,21 +141,12 @@ def __init__(self):
...
-        self.es = Elasticsearch([{'host': config['elastic_host'], 'port': config['elastic_port']}])
-
-        # delete index if exists
-        if self.es.indices.exists(config['index_name']):
-            print("Removing existing index " + config['index_name'])
-            self.es.indices.delete(index=config['index_name'])
-
-        # create index
-        self.es.indices.create(index=config['index_name'], ignore=400, body=settings)
+        client = MongoClient('mongodb://mongomcs:mongomcspassword@localhost:27017/mcs')
```
ThomasSchellenbergNextCentury, line 144: Do we care if this username / password is in the public codebase? Should it be read from an separate uncommitted config file?

eval1/create_json_ingest.py
```diff
@@ -234,7 +227,10 @@ def process_submission(self, filename):
 
                     self.object_id = self.object_id + 1
 
-        res = self.es.bulk(index=config['index_name'], body=bulk_data, refresh=True, request_timeout=30)
+        #res = self.es.bulk(index=config['index_name'], body=bulk_data, refresh=True, request_timeout=30)
```
ThomasSchellenbergNextCentury, line 230: I don't think you need this line anymore.

mcs_scene_ingest.py
```diff
@@ -41,9 +45,36 @@ def delete_keys_from_scene(scene, keys) -> dict:
...
+
+def ingest_to_mongo(index: str, ingest_files: dict):
+    collection = mongoDB[index]
+    result = collection.insert_many(ingest_files)
+    print(result)
+
+    keys = []
+    mydoc = collection.find()
+    for x in mydoc:
+        recursive_find_keys(x, keys, "")
```
ThomasSchellenbergNextCentury, line 69: What is `x`? Is it a search result?

eval1/create_json_ingest.py
```diff
@@ -30,6 +30,8 @@
 from elasticsearch import Elasticsearch
```
ThomasSchellenbergNextCentury, line 30: Do you still need the elasticsearch import here?

mcs_scene_ingest.py
brianpippin: so mydoc is the items in the collection, x is each item.

We create the keys collection at ingest time because this is very expensive to calculate and it allows the front end that wants to know the keys for creating searches to have a very quick query for them.

brianpippin: I will add that these variables could probably be named better.

eval1/create_json_ingest.py
rartiss55: Linking to convo here: https://github.com/NextCenturyCorporation/mcs-ui/pull/1/files/1258b330ba43801f37c6e4807d1024e1215f9d71#r534615432

PR #4, opened 2021-01-12
MCS-447: Update Ingest for Eval 3 Scenes/History

brianpippin:


agent_pair_scoring_scripty.py
```diff
@@ -0,0 +1,78 @@
+import argparse
```
rartiss55, line 1: Typo in script name at the end.

create_keys_script.py
```diff
@@ -0,0 +1,58 @@
+import argparse
```
rartiss55, line 1: Should we update the readme instructions for eval 3 data ingest/include these scripts?

mcs_scene_ingest.py
```diff
@@ -212,34 +240,54 @@ def ingest_history_files(folder: str, eval_name: str, performer: str, scene_fold
...
-            scene = load_json_file(scene_folder, history_item["name"] + "-debug.json")
+            scene = load_json_file(scene_folder, history_item["name"] + SCENE_DEBUG_EXTENSION)
             if scene:
-                if("observation" in scene):
-                    if(scene["observation"]):
-                        history_item["category"] = "observation"
-                        history_item["category_type"] = history_item["test_type"]
+                # For eval 3 going forward
+                if "test_type" not in history_item:
+                    history_item["scene_num"] = scene["sceneNumber"]
```
rartiss55, line 247: sceneNumber and sequenceNumber got swapped during ingest. I have another ticket to cover this + a few other things I accounted for the in the UI that we may want to fix with ingest later (MCS-520), but wondering if we should leave a comment in the code here for now/make a note to fix it for Eval 4 if we never fix it for this eval?

PR #5, opened 2021-02-23
MCS-520: Updates to previously ingested data

rartiss55:
Mainly, renamed and consolidated fields like sequenceNumber/scene_part_num/scene_num for clarity. Now it should be test_num and scene_num across all the data, and is stored as an integer regardless of eval (types were different before). Removed any duplicate fields like sequenceNumber, sceneNumber, etc. 

Also renamed sequenceId to hypercubeId to be consistent with newer scene files, and added the analysis UI url to the data in mcs_history as "url_string". 

To update existing data locally + on EC2, update_scene_num_refs.py needs to be run once. 

To view in the UI, see: https://github.com/NextCenturyCorporation/mcs-ui/pull/22

### DIFF
--- a/mcs_scene_ingest.py
+++ b/mcs_scene_ingest.py
@@ -142,11 +142,11 @@ def ingest_scene_files(folder: str, eval_name: str, performer: str) -> None:
             scene["test_num"] = int(scene["name"][-6:-2])
             scene["scene_num"] = int(scene["name"][-1:])
         else:
+            scene["scene_num"] = scene["sceneNumber"]
             if "sequenceNumber" in scene:
                 scene["test_num"] = scene["sequenceNumber"]
             else:
                 scene["test_num"] = scene["hypercubeNumber"]
-            scene["scene_num"] = scene["sceneNumber"]
             if "sequenceId" in scene["goal"]["sceneInfo"]:
                 scene["goal"]["sceneInfo"]["hypercubeId"] = scene["goal"]["sceneInfo"]["sequenceId"]
                 del scene["goal"]["sceneInfo"]["sequenceId"]
--- a/mcs_scene_ingest.py
+++ b/mcs_scene_ingest.py
@@ -143,13 +143,16 @@ def ingest_scene_files(folder: str, eval_name: str, performer: str) -> None:
             scene["scene_num"] = int(scene["name"][-1:])
         else:
             scene["scene_num"] = scene["sceneNumber"]
+
             if "sequenceNumber" in scene:
                 scene["test_num"] = scene["sequenceNumber"]
             else:
                 scene["test_num"] = scene["hypercubeNumber"]
+
             if "sequenceId" in scene["goal"]["sceneInfo"]:
                 scene["goal"]["sceneInfo"]["hypercubeId"] = scene["goal"]["sceneInfo"]["sequenceId"]
                 del scene["goal"]["sceneInfo"]["sequenceId"]
+
         scene = delete_keys_from_scene(scene, KEYS_TO_DELETE)
         ingest_scenes.append(scene)
 
--- a/mcs_scene_ingest.py
+++ b/mcs_scene_ingest.py
@@ -182,8 +182,6 @@ def ingest_history_files(folder: str, eval_name: str, performer: str, scene_fold
             history_item["test_type"] = history_item["name"][:-7]
             history_item["test_num"] = int(history_item["name"][-6:-2])
             history_item["scene_num"] = int(history_item["name"][-1:])
-            history_item["url_string"] = ("eval=" + history_item["eval"] + "&test_type=" + history_item["test_type"] +
-                                         "&test_num=" + str(history_item["test_num"]) + "&scene=" + str(history_item["scene_num"]))
         else: 
             history_item["performer"] = TEAM_MAPPING_DICT[history["info"]["team"]]
             history_item["name"] = history["info"]["name"]
@@ -265,8 +263,6 @@ def ingest_history_files(folder: str, eval_name: str, performer: str, scene_fold
                         history_item["category_type"] = scene["goal"]["sceneInfo"]["name"][:-3]
                     else:
                         history_item["category_type"] = scene["goal"]["sceneInfo"]["tertiaryType"]
-                    history_item["url_string"] = ("eval=" + history_item["eval"] + "&category_type=" + history_item["category_type"] +
-                        "&test_num=" + str(history_item["test_num"]) + "&scene=" + str(history_item["scene_num"]))
                 # For eval 2 
                 else:
                     if("observation" in scene):
--- a/mcs_scene_ingest.py
+++ b/mcs_scene_ingest.py
@@ -147,6 +147,9 @@ def ingest_scene_files(folder: str, eval_name: str, performer: str) -> None:
             else:
                 scene["test_num"] = scene["hypercubeNumber"]
             scene["scene_num"] = scene["sceneNumber"]
+            if "sequenceId" in scene["goal"]["sceneInfo"]:
+                scene["goal"]["sceneInfo"]["hypercubeId"] = scene["goal"]["sceneInfo"]["sequenceId"]
+                del scene["goal"]["sceneInfo"]["sequenceId"]
         scene = delete_keys_from_scene(scene, KEYS_TO_DELETE)
         ingest_scenes.append(scene)
 
--- a/mcs_scene_ingest.py
+++ b/mcs_scene_ingest.py
@@ -139,9 +139,14 @@ def ingest_scene_files(folder: str, eval_name: str, performer: str) -> None:
         scene["performer"] = performer
         if "2" in eval_name:
             scene["test_type"] = scene["name"][:-7]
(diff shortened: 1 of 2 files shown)


mcs_scene_ingest.py
```diff
@@ -361,7 +372,8 @@ def ingest_history_files(folder: str, eval_name: str, performer: str, scene_fold
                     else:
                         history_item["score"]["weighted_score"] = history_item["score"]["score"]
                         history_item["score"]["weighted_score_worth"] = 1
-                        history_item["score"]["weighted_confidence"] = history_item["score"]["confidence"]
+                        if "confidence" in history_item["score"]:
```
rartiss55, line 375: This and the change on line 389 were to handle a couple of errors that were thrown when I attempted to test and reingest eval 2 data.

update_scene_num_refs.py
```diff
@@ -0,0 +1,198 @@
...
+
+def update_scene_num_refs_history_eval_3():
+    print("Begin Processing " + EVAL_3_RESULTS)
+    collection = mongoDB[HISTORY_INDEX]
+
+    # rename scene_part_num field to test_num
+    result = collection.update_many({"eval": EVAL_3_RESULTS, "scene_part_num": {"$exists": True}}, {"$rename": {'scene_part_num' : 'test_num'}}, False)
+    print("update_many performed on " + str(result.matched_count) + " documents")
+
+    categories = [
```
rartiss55, line 32: Splitting up the updates to mcs_history by category because my script kept hanging. In hindsight it might've made more sense to just write a straight mongo shell script instead of going the pymongo route.

update_scene_num_refs.py
```diff
@@ -0,0 +1,198 @@
...
+# We might want to move mongo user/pass to new file
+client = MongoClient('mongodb://mongomcs:mongomcspassword@localhost:27017/mcs')
+mongoDB = client['mcs']
+
+def update_scene_num_refs_history_eval_3():
+    print("Begin Processing " + EVAL_3_RESULTS)
+    collection = mongoDB[HISTORY_INDEX]
+
+    # rename scene_part_num field to test_num
+    result = collection.update_many({"eval": EVAL_3_RESULTS, "scene_part_num": {"$exists": True}}, {"$rename": {'scene_part_num' : 'test_num'}}, False)
```
rartiss55, line 29: update_many here also helped things move a bit faster since part of this is just renaming an existing field.

update_scene_num_refs.py
```diff
@@ -0,0 +1,198 @@
...
+        if "sequenceNumber" in doc["keys"]:
+            keys_to_remove.append("sequenceNumber")
+        if "sceneNumber" in doc["keys"]:
+            keys_to_remove.append("sceneNumber")
+
+        result = collection.update_one({"_id": ObjectId(doc["_id"])}, {"$push": {'keys': { "$each": keys_to_add}}}, True)
+        result = collection.update_one({"_id": ObjectId(doc["_id"])}, {"$pull": {'keys': { "$in": keys_to_remove}}}, True)
+   
+def delete_old_saved_query():
+    print("Deleting single saved query...")
```
rartiss55, line 180: I checked, and the only saved query referencing an outdated field is one of my test queries, so I just went ahead and deleted it.

update_scene_num_refs.py
```diff
@@ -0,0 +1,198 @@
...
+    results = collection.delete_one({"_id": ObjectId("60008951c9ef6d5e7a77eb65")})
+
+# one time run script (MCS-520)
+def main():
+    update_scene_num_refs_history_eval_2()
+    update_scene_num_refs_history_eval_3()
+    update_scene_num_refs_scenes_eval_2()
+    update_scene_num_refs_scenes_eval_3()
+
+    update_collection_keys()
```
rartiss55, line 192: Needed to update key collections as well.

mcs_scene_ingest.py
```diff
@@ -172,9 +180,10 @@ def ingest_history_files(folder: str, eval_name: str, performer: str, scene_fold
             history_item["name"] = get_scene_name_from_history_text_file(file, "-202.+-")
             SCENE_DEBUG_EXTENSION = "-debug.json"
             history_item["test_type"] = history_item["name"][:-7]
-            history_item["scene_num"] = history_item["name"][-6:-2]
-            history_item["scene_part_num"] = history_item["name"][-1:]
-            history_item["url_string"] = "test_type=" + history_item["test_type"] + "&scene_num=" + history_item["scene_num"] + "&scene_part_num=" + history_item["scene_part_num"] + "&performer=" + history_item["performer"]
+            history_item["test_num"] = int(history_item["name"][-6:-2])
+            history_item["scene_num"] = int(history_item["name"][-1:])
+            history_item["url_string"] = ("eval=" + history_item["eval"] + "&test_type=" + history_item["test_type"] +
```
brianpippin, line 185: I think we should be removing the url string field, it's not part of the data and was a leftover from needing to have a way from neon to get to analysis.  The ui should build the string and it shouldn't be part of the database.

update_scene_num_refs.py
```diff
@@ -0,0 +1,198 @@
...
+        "agents object preference",
+        "object permanence",
+        "retrieval_container",
+        "retrieval_obstacle",
+        "retrieval_occluder",
+        "shape constancy",
+        "spatio temporal continuity"
+    ]
+    
+    # update url_string one by one
```
brianpippin, line 44: See above comment, I'd like to remove url_string from the database and have the UI build it's one string

update_scene_num_refs.py
```diff
@@ -0,0 +1,198 @@
...
+    collection = mongoDB[HISTORY_INDEX]
+
+    documents = list(collection.find({"eval": EVAL_2_RESULTS, "scene_num" : {"$exists" : True}, "scene_part_num": {"$exists": True}}).batch_size(1000))
+    print("Found " + str(len(documents)) + " results")
+
+    for doc in documents:
+
+        test_num = int(doc["scene_num"])
+        scene = int(doc["scene_part_num"])
+        url_string = "eval=" + doc["eval"] + "&test_type=" + doc["test_type"] + "&test_num=" + str(test_num) + "&scene=" + str(scene)
```
brianpippin, line 66: Remove url_string

brianpippin:
I need to check this out and run the update script.  I do think though we should not worry about updating ingest to necessarily be backwards compatible at this point.  We have mongo backups currently and I created a ticket to backup the database nightly so I don't think we will ever need to reingest data at this point.  We should only focus on ingesting new evaluations and creating scripts when necessary to update the mongo db.  Let me know if there are any concerns with that approach?

mcs_scene_ingest.py
rartiss55: Okay, saw the eval 2 field and thought it'd be neat but you can make the argument either way. I took it out (also makes the one time script run a lot faster).

update_scene_num_refs.py
rartiss55: done

update_scene_num_refs.py
rartiss55: done

rartiss55:
> I need to check this out and run the update script. I do think though we should not worry about updating ingest to necessarily be backwards compatible at this point. We have mongo backups currently and I created a ticket to backup the database nightly so I don't think we will ever need to reingest data at this point. We should only focus on ingesting new evaluations and creating scripts when necessary to update the mongo db. Let me know if there are any concerns with that approach?

I think that sounds good going forward and reduces some of the testing time for us making these types of changes. To clarify, we shouldn't ever need to update eval 2/anything prior to that but probably want to keep eval 3+ ingest synced with whatever changes we make going forward, right?

mcs_scene_ingest.py
```diff
@@ -139,9 +139,17 @@ def ingest_scene_files(folder: str, eval_name: str, performer: str) -> None:
...
-            scene["scene_part_num"] = scene["name"][-1:]
-
+            scene["test_num"] = int(scene["name"][-6:-2])
+            scene["scene_num"] = int(scene["name"][-1:])
+        else:
+            if "sequenceNumber" in scene:
+                scene["test_num"] = scene["sequenceNumber"]
+            else:
+                scene["test_num"] = scene["hypercubeNumber"]
+            scene["scene_num"] = scene["sceneNumber"]
```
brianpippin, line 149: Indentation here.

update_scene_num_refs.py
brianpippin: I think that all but collection_keys is OBE now right?  I would need to check the code but I thought we had refactored to only use collection keys.

brianpippin (approved):
I think in the future if we can just write mongo update scripts we probably don't need to write them in python, but this all looks good to me.

update_scene_num_refs.py
rartiss55: Ah okay, so I guess some of that got missed in server.schema.js. I made a ticket here: https://nextcentury.atlassian.net/browse/MCS-583

mcs_scene_ingest.py
rartiss55: I added some spaces/moved that line for clarity.

PR #7, opened 2021-03-09
MCS 539: Add script to automate single history file ingestion

brianpippin:
Aand update ingest to check for scene in mongo

### DIFF
--- a/mcs_scene_ingest.py
+++ b/mcs_scene_ingest.py
@@ -80,22 +80,6 @@ def ingest_to_mongo(index: str, ingest_files: dict):
     result = collection.insert_many(ingest_files)
     print(result)
 
-    # ----- Moving to create_keys_script that will run after ingestion -----
-    # Loop through documents to generate a keys collection to help
-    #   speed in loading keys in UI
-    # keys = []
-    # documents = collection.find()
-    # for doc in documents:
-    #     recursive_find_keys(doc, keys, "")
-
-    # keys_dict = {}
-    # keys_dict["keys"] = keys
-
-    # collection = mongoDB[index + "_keys"]
-    # collection.drop()
-    # result = collection.insert_one(keys_dict)
-    # print(result)
-
 
 def find_scene_files(folder: str) -> dict:
     scene_files = [f for f in os.listdir(folder) if str(f).endswith("debug.json")]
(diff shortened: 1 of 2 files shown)


mcs_scene_ingest.py
```diff
@@ -160,73 +144,69 @@ def ingest_scene_files(folder: str, eval_name: str, performer: str) -> None:
     ingest_to_mongo(SCENE_INDEX, ingest_scenes)
 
 
-def ingest_history_files(folder: str, eval_name: str, performer: str, scene_folder: str, extension: str) -> None:
-    history_files = find_history_files(folder, extension)
-    ingest_history = []
-    SCENE_DEBUG_EXTENSION = "_debug.json"
+def automated_history_ingest_file(history_file: str, folder: str):
```
brianpippin, line 147: New Function called from mcs_automated_ingest.py

mcs_scene_ingest.py
```diff
@@ -385,21 +382,29 @@ def ingest_history_files(folder: str, eval_name: str, performer: str, scene_fold
...
-            else:
-                ingest_history[replacementIndex] = history_item
+            
+    # Check for duplicate Mess History files that don't include any steps
+    if steps:
+        history_item["steps"] = steps
+
+    return history_item
+
+def ingest_history_files(folder: str, eval_name: str, performer: str, scene_folder: str, extension: str) -> None:
```
brianpippin, line 392: Most of the code that was in this function was just moved to:  build_history_item function

mcs_scene_ingest.py
```diff
@@ -160,73 +144,69 @@ def ingest_scene_files(folder: str, eval_name: str, performer: str) -> None:
...
+                collection.replace_one({"_id": item["_id"]}, history_item)
 
-        # Legacy Eval 2 History files will be txt files and not json
-        if extension == 'txt':
-            history = load_history_text_file(folder, file)
-        else:
-            history = load_json_file(folder, file)
 
-        history_item = {}
+def build_history_item(history_file: str, folder: str, eval_name: str, performer: str, scene_folder: str, extension: str) -> dict:
```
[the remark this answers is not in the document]
brianpippin: Most of the code in this file is unchanged except where noted.

mcs_scene_ingest.py
```diff
@@ -241,141 +221,158 @@ def ingest_history_files(folder: str, eval_name: str, performer: str, scene_fold
...
+                    interactive_goal_achieved = 1
+            new_step["output"] = output
+            steps.append(new_step)
+
+        history_item["score"] = history["score"]
+    
+    history_item["step_counter"] = number_steps         
+
+    scene = None
+    if scene_folder is None:
```
brianpippin, line 261: New function to load scene data from mongo if a scene folder is not passed in via ingest command.

mcs_scene_ingest.py
```diff
@@ -385,21 +382,29 @@ def ingest_history_files(folder: str, eval_name: str, performer: str, scene_fold
...
                     history_item["score"]["weighted_score_worth"] = 1
                     history_item["score"]["weighted_confidence"] = history_item["score"]["confidence"]
-                
-        print(history_item["score"])
-        # Check for duplicate Mess History files that don't include any steps
-        if steps:
-            history_item["steps"] = steps
-
-            replacementIndex = -1
-            for index, item in enumerate(ingest_history):
```
brianpippin, line 395: This code is still in ingest_history_files function.

mcs_scene_ingest.py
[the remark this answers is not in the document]
deanwetherby: Can you think of a way to break this function down into smaller pieces? It seems like each area you explain with code could be another testable sub-function.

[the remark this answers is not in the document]
brianpippin: That's the same function that's been ingesting since Eval2 just moved.  I would love to break it up, but I would like to wait till after Eval3.5 so we don't introduce any potential issues with 3.5 with the quick turn around.  I just don't know if I have enough time to fully test all the different vectors before then.  

100% agree it should be broken up and then individual tests could be written for each component.

deanwetherby (approved):

PR #11, opened 2021-03-31
MCS-586: Ingest Refactor

brianpippin:


brianpippin:
Lots of changes for flake8, refactoring larger functions into smaller functions and added tests for the trickier functions.

mcs_scene_ingest.py
```diff
@@ -170,259 +153,259 @@ def ingest_scene_files(folder: str, eval_name: str) -> None:
...
+    return (new_step, interactive_reward, interactive_goal_achieved)
+
+
+def add_weighted_cube_scoring(history_item: dict, scene: dict) -> tuple:
+    weighted_score, weighted_score_worth, weighted_confidence = 0, 0, 0
+    if "goal" in scene:
+        if "sceneInfo" in scene["goal"]:
+            if scene["goal"]["sceneInfo"]["tertiaryType"] == "shape constancy":
+                if history_item[
+                        "scene_goal_id"] in SHAPE_CONSTANCY_DUPLICATE_CUBE:
```
rartiss55, line 217: Do we need to update this at all for Eval 4 (will the ids change for instance)?

requirements.txt
```diff
@@ -0,0 +1 @@
+pymongo==3.11.3
```
rartiss55, line 1: Should add flake8 here as well.

tests/test_mcs_scene_ingest.py
```diff
@@ -0,0 +1,71 @@
+import unittest
```
rartiss55, line 1: Yay tests! Should we add how to run/"python3 -m unittest" to the README?

scripts/update_scene_num_refs.py
```diff
@@ -1,13 +1,5 @@
-import argparse
-import cmd
-import os
-import io
-import json
-
-from collections.abc import MutableMapping 
 from pymongo import MongoClient
```
rartiss55, line 1: I know this was a one off script I wrote a little while ago, and I do think there's value in keeping stuff like that around for tracking, but wondering if we should label them better, maybe put them in a separate folder?

mcs_scene_ingest.py
```diff
@@ -170,259 +153,259 @@ def ingest_scene_files(folder: str, eval_name: str) -> None:
...
-            history_item["score"]["ground_truth"] = 1
+        # Add some basic scene information into history object to make
+        #    UI load times faster then having to query scene every time
+        history_item["scene_num"] = scene["scene_num"]
+        history_item["test_num"] = scene["test_num"]
+        history_item["scene_goal_id"] = scene["goal"]["sceneInfo"]["id"][0]
+        history_item["test_type"] = scene["goal"]["sceneInfo"]["secondaryType"]
+        history_item["category"] = scene["goal"]["sceneInfo"]["primaryType"]
+
+        # TODO: Make sure this was fixed so we no longer need to do
```
rartiss55, line 351: The ticket is MCS-578 if you think it'd be good to reference that here. 

I was looking at the data and I think we might actually need secondaryType + "_" + tertiaryType for retrieval tasks, but I added a note about that in the ticket.

mcs_scene_ingest.py
```diff
@@ -170,259 +153,259 @@ def ingest_scene_files(folder: str, eval_name: str) -> None:
...
+            interactive_reward,
+            interactive_goal_achieved
+        ) = build_new_step_obj(
+            step,
+            interactive_reward,
+            interactive_goal_achieved,
+            number_steps)
+        steps.append(new_step)
+
+    history_item["step_counter"] = number_steps
```
rartiss55, line 331: Did some end to end testing with ingestion and the UI, and I think the steps actually need to be added to the history item here:
```
history_item["steps"] = steps
```

mcs_scene_ingest.py
```diff
@@ -170,259 +153,259 @@ def ingest_scene_files(folder: str, eval_name: str) -> None:
...
 
     for file in history_files:
-        history_item = build_history_item(file, folder, eval_name, performer, scene_folder, extension)
+        history_item = build_history_item(
+            file, folder, eval_name, performer, scene_folder, extension)
 
         replacementIndex = -1
         for index, item in enumerate(ingest_history):
-            if item["filename"] == history_item["filename"] and history_item["fileTimestamp"] > item["fileTimestamp"]:
+            if item["filename"] == history_item["filename"] and history_item[
```
rartiss55, line 407: I think you took out the "filename" field (which is fine, seemed unnecessary), but I got a key error here.

mcs_scene_ingest.py
```diff
@@ -170,259 +153,259 @@ def ingest_scene_files(folder: str, eval_name: str) -> None:
...
-            if "score" not in history_item:
-                history_item["score"] = {}
-                history_item["score"]["classification"] = "end"
-                history_item["score"]["confidence"] = 0
-            history_item["score"]["score"] = interactive_goal_achieved
-            history_item["score"]["reward"] = interactive_reward
-            history_item["score"]["ground_truth"] = 1
+        # Add some basic scene information into history object to make
+        #    UI load times faster then having to query scene every time
+        history_item["scene_num"] = scene["scene_num"]
```
rartiss55, line 345: I get a key error here when I try to run with the --scene-folder argument. I think that's okay since we won't be using that anymore, but we should either take out that flag or fix the error.

scripts/agent_pair_scoring_script.py
```diff
@@ -16,15 +8,26 @@
...
+    "IBM-MIT-Harvard-Stanford",
+    "MESS-UCBerkeley",
+    "OPICS (OSU, UU, NYU)"
+]
+
+agent_tasks = [
+    "agents efficient action path lure",
+    "agents efficient action time control",
+    "agents object preference"
+]
```
ThomasSchellenbergNextCentury, line 21: You'll need to update this list with the Eval 4 agent tasks, once we have the full list (might be worth making a ticket for it).

mcs_scene_ingest.py
brianpippin: I asked Koleen about what was weighted for Eval 4, and she said only shape constancy, so that's why I removed the others.  We only need to update the IDs in the array at the top if it is different but it was my understanding it wouldn't be.

mcs_scene_ingest.py
brianpippin: Old line 412 is where that is at, that is where it should be.  Adding it back in!

mcs_scene_ingest.py
brianpippin: I still want to keep that possibility for now.  I thought I tested with that argument, I will double check it is working.

brianpippin: When you got that error, did you make sure the scene file for the history file you were running was actually in the scene folder?

mcs_scene_ingest.py
brianpippin: Added a reference to the ticket

mcs_scene_ingest.py
brianpippin: Updated to fullFilename

scripts/agent_pair_scoring_script.py
```diff
@@ -16,15 +8,26 @@
 client = MongoClient('mongodb://mongomcs:mongomcspassword@localhost:27017/mcs')
 mongoDB = client['mcs']
 
-performers = [ "IBM-MIT-Harvard-Stanford", "MESS-UCBerkeley", "OPICS (OSU, UU, NYU)" ]
-agent_tasks = ["agents efficient action path lure", "agents efficient action time control", "agents object preference"]
+performers = [
+    "IBM-MIT-Harvard-Stanford",
```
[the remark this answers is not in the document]
brianpippin: This is a script we saved for running some update, we shouldn't necessarily be running again, unless we need to modify and run it again.

scripts/update_scene_num_refs.py
brianpippin: That's why I moved them to the scripts folder, as most of the scripts in there will be one off scripts we need to run, and then save for reference.

mcs_scene_ingest.py
rartiss55: Ah okay, I see the issue. When running from a scene folder instead of a mongo record, it needs to be something like this since the keys are different in the scene itself for the scene_num/test_num fields:
```
        if scene_folder is None:
            history_item["scene_num"] = scene["scene_num"]
            history_item["test_num"] = scene["test_num"]
        else:
            history_item["scene_num"] = scene["sceneNumber"]
            if "sequenceNumber" in scene:
                history_item["test_num"] = scene["sequenceNumber"]
            else: 
                history_item["test_num"] = scene["hypercubeNumber"]
```

I think I had just the scene_folder case before, and that got taken out because it didn't work right when not running from a scene folder (which makes sense why that wouldn't work).

rartiss55 (approved):

PR #12, opened 2021-04-21
Mcs 633 fps to ingest

kdrumm:


README.md
```diff
@@ -1,11 +1,49 @@
...
+```
+python mcs_scene_ingest.py --folder ../MCS/SCENE_HISTORY/  --eval_name eval-test --type history    --performer "baseline"
+```
+
+# Recommendations when testing:
+
+* When creating scenes, you ingest the debug files (move them to their own directory)
+* When testing, use an eval name that is unique, doesn't look like a real eval, and is easy to notice.
+* Make sure you use same eval name for scenes and history when ingesting
+* Verify info.team is set in history json (normally set in config when running eval)
```
brianpippin, line 46: I would also add something about using a metadata level, probably oracle.

brianpippin (approved):

README.md
kdrumm: Good idea.

brianpippin (approved):

PR #13, opened 2021-04-22
MCS-583:  Removes old keys collections, framework for running database update scripts

brianpippin:
This has a companion ticket in mcs-ui, MCS-583 there as well, that actually removes the use of the old keys in the codebase.


deployment_script.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+def check_version():
+    collection = mongoDB[VERSION_COLLECTION]
+    version_obj = collection.find_one()
+    # return true if it is a newer db version
+    if db_version > version_obj['version']:
+        return True
+    else:
+        return False
```
ThomasSchellenbergNextCentury, line 22: Not a big deal, but could just be: `return db_version > version_obj['version']`

scripts/drop_extra_keys_collection.py
```diff
@@ -0,0 +1,11 @@
+from pymongo import MongoClient
+
+client = MongoClient(
+    'mongodb://mongomcs:mongomcspassword@localhost:27017/mcs')
```
ThomasSchellenbergNextCentury, line 4: The mongo client seems to be used in multiple different python files. Should it (either the string, or the MongoClient object) be created once in a separate python file, and then imported into all of these other python files? That way, if you need to change the host/password in the future, you do it in one place.

kdrumm: Agreed with Thomas.  Additionally, is this file necessary?  It seems we could easily move the small function below to the file it is called and solve the above problem.  Not really a big deal and I could easily be convinced as a proponent of more, smaller files.

brianpippin: This file is necessary because I am trying to set up a paradigm for running deployment scripts, sometimes the scripts with be larger sometimes not, but if you put the actual code in the deployment script, then every time you modify that file for the next mongo script, we'd lose the code to only git history, this way we can keep a record of scripts.  Perhaps we should add the version to the filename as well to make that easier to map back.

kdrumm: Makes sense to me.  Thanks.

kdrumm (approved):

deployment_script.py
```diff
@@ -0,0 +1,40 @@
+from pymongo import MongoClient
+import scripts.drop_extra_keys_collection as drop_extra_keys_collection
```
rartiss55, line 2: This is neat! I tested this out locally and only thing is this import reference didn't really work. I think it might have to do with the script name having periods in it (0.4.1_drop_extra_keys_collection.py). If we want to have version numbers in the titles (which I think is nice) maybe we could do 0-4-1-scriptname instead or something like that.

mcs_scene_ingest.py
```diff
@@ -76,7 +76,8 @@ def delete_keys_from_scene(scene, keys) -> dict:
 def ingest_to_mongo(index: str, ingest_files: dict):
     collection = mongoDB[index]
     result = collection.insert_many(ingest_files)
-    print("Inserted {0} out of {1}.  Result:".format(len(result.inserted_ids), len(ingest_files), result))
+    print("Inserted {0} out of {1}.  Result: {2}".format(
```
deanwetherby, line 79: Could use f-strings here although `.format` works too.
```
print("Inserted {len(result.inserted_ids)} out of {len(ingest_files)}. Result: {result}"
```

ThomasSchellenbergNextCentury (approved):
The code LGTM

PR #15, opened 2021-05-26
Fixes for various tickets (MCS-599, MCS-578, MCS-646)

rartiss55:
A bunch of small fixes: Fixing category_type field for history records, adding delta_time field to ingest, updating constants for eval 3.5 rerun.

### DIFF
--- a/mcs_scene_ingest.py
+++ b/mcs_scene_ingest.py
@@ -358,12 +358,10 @@ def build_history_item(
         history_item["test_type"] = scene["goal"]["sceneInfo"]["secondaryType"]
         history_item["category"] = scene["goal"]["sceneInfo"]["primaryType"]
 
-        # TODO: Make sure this was fixed so we no longer need to do
-        #   check, might need quaternary type
-        # MCS-578 https://nextcentury.atlassian.net/jira/software/projects/MCS/boards/94?selectedIssue=MCS-578&text=578 # noqa: E501
-        if scene["goal"]["sceneInfo"]["tertiaryType"] == "retrieval":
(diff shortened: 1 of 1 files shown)


rartiss55:
Did some testing and noticed this line: 
https://github.com/NextCenturyCorporation/mcs-ingest/blob/42d6e75b666dd40cc1ad9ab4d6cf42a577f5cfcd/mcs_scene_ingest.py#L343

Wondering if we should add "eval" as well to that check in case we reuse file names for future evals (like if we call another file "delta_0001_01" in the future)?

brianpippin (approved):
This code looks good, and I trust you on the secondary/tertiary fix, i assume you ingested a new file and made sure it came through okay?

deanwetherby (approved):

ThomasSchellenbergNextCentury (approved):
The code looks good.

rartiss55:
> This code looks good, and I trust you on the secondary/tertiary fix, i assume you ingested a new file and made sure it came through okay?

Yup!

rartiss55:
> Did some testing and noticed this line:
> https://github.com/NextCenturyCorporation/mcs-ingest/blob/42d6e75b666dd40cc1ad9ab4d6cf42a577f5cfcd/mcs_scene_ingest.py#L343
> 
> Wondering if we should add "eval" as well to that check in case we reuse file names for future evals (like if we call another scene file "delta_0001_01" in the future)?

My latest commit addresses this -- added an additional dict since history records have the eval name "evaluation x results" and scene records use "evaluation x scenes"

brianpippin (approved):

PR #16, opened 2021-06-08
MCS-614:  Automated Key Generation

brianpippin:
Once the first element of a new evaluation is added to the database, auto generate the keys for that new entry.


scripts/create_keys_script.py
```diff
@@ -43,17 +33,3 @@ def find_collection_keys(index: str, collection_name: str):
     result = collection.update_one(
         {"name": collection_name}, {"$set": keys_dict}, True)
     print(result)
-
-
-def main():
```
ThomasSchellenbergNextCentury, line 48: Now that you've removed the "main" function from this file, and it's more like a utility file, should it be renamed to remove the `_script` suffix? To me, "_script" implies that the file is runnable on its own (but maybe that's just me?).

brianpippin: Thomas I updated this and moved it out of the script folder and renamed the file.

ThomasSchellenbergNextCentury (approved):
I didn't try running it, but the code LGTM

mcs_scene_ingest.py
```diff
@@ -419,6 +436,13 @@ def automated_history_ingest_file(history_file: str, folder: str) -> None:
                 print(f"Updating {history_item['name']}")
                 collection.replace_one({"_id": item["_id"]}, history_item)
 
+    # Add Keys when a new evluation item is created
+    collection_count = collection.find(
+        {"eval": history_item["eval"]}).count()
+    if collection_count == 1:
+        create_collection_keys.find_collection_keys(
```
rartiss55, line 443: Looks good -- I like that this is automated now! The only thing I'd say is that I like to test locally and didn't see a new line in collection_keys, but I think that's because this is only called for the automated process? I think that's fine but maybe we can note that more explicitly somewhere.

tests/test_mcs_scene_ingest.py
```diff
@@ -49,14 +49,14 @@ def test_build_scene_item(self):
         self.assertEqual(scene.get("sequenceNumber"), None)
         self.assertEqual(scene.get("hypercubeNumber"), None)
 
-    def test_determine_evaluation_name(self):
-        eval_name = mcs_scene_ingest.determine_evaluation_name(
+    def test_ddetermine_evaluation_hist_name(self):
```
rartiss55, line 52: Typo here.

mcs_scene_ingest.py
brianpippin: Pushed a fix for legacy way just to have it in case.

tests/test_mcs_scene_ingest.py
brianpippin: Good catch, fixed.

mcs_scene_ingest.py
```diff
@@ -471,6 +474,9 @@ def ingest_history_files(
 
     ingest_to_mongo(HISTORY_INDEX, ingest_history)
 
+    create_collection_keys.find_collection_keys(
+            SCENE_INDEX, eval_name, mongoDB)
```
rartiss55, line 478: I think this one needs to be HISTORY_INDEX.

brianpippin: Ugh yes it does.

brianpippin: Fixed and pushed.

rartiss55 (approved):

PR #17, opened 2021-06-22
[Ingest] MCS-700 Consolidation of Debug Scene File Properties

ThomasSchellenbergNextCentury:
- Updated the scene ingest code to read from the new debug scene file properties.
- Added some mongo setup instructions to the README.

You can generate example scenes on your own, but here's a set that I used for my testing: https://ta2-examples.s3.amazonaws.com/mcs700_scenes.zip

Corresponding Scene Generator PR: https://github.com/NextCenturyCorporation/mcs-private/pull/32
Corresponding Python PR: https://github.com/NextCenturyCorporation/MCS/pull/357

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,20 @@
 # mcs-ingest
-Ingest data into Elasticsearch for MCS
 
-# Setup
+Ingest data into MongoDB for MCS
+
+# Mongo Setup
+
+## Standalone
+
+See the official MongoDB installation page here: https://docs.mongodb.com/manual/installation/
+
+Change each instantiation of the MongoClient in our Python code to: `MongoClient('mongodb://localhost:27017/mcs')`
+
+## With the MCS UI
+
+See the mcs-ui README here: https://github.com/NextCenturyCorporation/mcs-ui/blob/master/README.md
+
+# Python Setup
 
 ## Using python virtual environment:
 
--- a/mcs_scene_ingest.py
+++ b/mcs_scene_ingest.py
@@ -18,9 +18,7 @@ mongoDB = client['mcs']
 #    that we want to just map the fields we want to the schema
 KEYS_TO_DELETE = [
     'image',
-    'sequenceNumber',
-    'hypercubeNumber',
-    'sceneNumber'
+    'debug'
 ]
 
 SCENE_INDEX = "mcs_scenes"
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -1,7 +1,20 @@
...
+
+## Standalone
+
+See the official MongoDB installation page here: https://docs.mongodb.com/manual/installation/
+
+Change each instantiation of the MongoClient in our Python code to: `MongoClient('mongodb://localhost:27017/mcs')`
+
+## With the MCS UI
+
+See the mcs-ui README here: https://github.com/NextCenturyCorporation/mcs-ui/blob/master/README.md
```
ThomasSchellenbergNextCentury, line 15: I ended up just installing Mongo standalone, since I didn't need to test the UI. Feel free to expand on these instructions as you think is best.

mcs_scene_ingest.py
```diff
@@ -302,7 +300,7 @@ def process_score(
             history_item["score"] = {}
             history_item["score"]["score"] = -1
             history_item["score"]["ground_truth"] = 1 if "plausible" == scene[
-                "answer"]["choice"] else 0
+                "goal"]["answer"]["choice"] else 0
```
ThomasSchellenbergNextCentury, line 303: This wasn't related to my ticket, but looked like a bug, and seemed easy to fix.

brianpippin: I believe this was a legacy pathway, where if there was no score in history_item, then the scene didn't have a goal answer choice, but just answer choice, I could be wrong though.  We might only run into this bug with older scene/history files, so the whole thing might be OBE, maybe Rachel remembers as well.

rartiss55: Yeah this part looks like it applies to the old pre-Eval 3 stuff.

brianpippin (approved):

rartiss55:
This looks fine, but I'd hold off on merging until after we rerun 3.5 (I think this wouldn't work right with the 3.5/3.75 files since those scene files don't have a debug section, right?

rartiss55:
@ThomasSchellenbergNextCentury Brian set up a development branch so we don't have to be in a holding pattern for eval 3.75 vs 4 -- I went ahead and updated the base branch for your PR, so you should be good.

rartiss55 (approved):

mcs_scene_ingest.py
ThomasSchellenbergNextCentury: Done. I also added a comment for future reference.

PR #22, opened 2021-08-13
Mcs 580 scorecard

cedorman:


requirements.txt
```diff
@@ -1,2 +1,5 @@
 pymongo==3.11.3
-flake8==3.8.4
\ No newline at end of file
+flake8==3.8.4
+numpy==1.19.4
+pandas==1.2.3
+machine_common_sense==0.4.1.1
```
kdrumm, line 5: We are on 0.4.3 or at least something new than this right?  Why are we using an old version?

scorecard/generator/path_plotter.py
```diff
@@ -0,0 +1,242 @@
...
+            if not obj.get('objectId', '').startswith('ceiling') and not
+            obj.get('objectId', '').startswith('floor')
+        ]
+        objects = scene_event.metadata.get('objects', [])
+        return filtered_structural_objects + objects
+
+    def _initialize_plot(self, step_number: int):
+        """Create the plot"""
+        plt.xlim(self.MINIMUM_ROOM_DIMENSION, self.MAXIMUM_ROOM_DIMENSION)
+        plt.ylim(self.MINIMUM_ROOM_DIMENSION, self.MAXIMUM_ROOM_DIMENSION)
```
kdrumm, line 96: Room dimensions are not fix anymore.  We will need to get this from some output.

scorecard/utils.py
```diff
@@ -0,0 +1,19 @@
+#
+# Useful utility functions
+#
+
+def minAngDist(a, b):
```
kdrumm, line 5: Recently we've been trying to stay away from util packages/classes because of the vague-ness.  @deanwetherby for the details/links to why.  This could simply be moved to a more descriptive file like Math or anglemath or something like that.  (naming things is hard)

scorecard_generate_revisit_data.py
```diff
@@ -0,0 +1,121 @@
...
+    return None, None
+
+
+class FakeEvent:
+    """the plotter wants an ai2thor.event which contains metadata,
+    but we don't have one, so create a FakeEvent object that contains
+    metadata and pass that"""
+
+    def __init__(self, metadata):
+        self.metadata = metadata
```
kdrumm, line 49: Can we just import the FakeEvent in 'create_path_for_existing.py' (or vice versa)

scorecard/generator/path_plotter.py
```diff
@@ -0,0 +1,242 @@
+#
+# path_plotter plots the path of an actor in the room, so we can visualize it
+#
+# It is based on machine_common_sense.plotter.py
+#
```
ThomasSchellenbergNextCentury, line 5: What's the difference between this class and the MCS plotter? Is this class just a copy?

scorecard/scorecard.py
```diff
@@ -0,0 +1,191 @@
...
+import json
+
+import numpy as np
+import pandas
+
+from scorecard.utils import minAngDist
+
+# Assume that the entire room space goes from -5 to 5 in X and Z.  If not,
+# then will need to change this line or read in from scene json file.
+SPACE_SIZE = 5.
```
ThomasSchellenbergNextCentury, line 14: Please note that the room size will be different for both interactive object permanence and reorientation scenes in Eval 4.

scorecard/scorecard.py
```diff
@@ -0,0 +1,191 @@
...
+class GridHistory:
+    """A history of the times a grid square has been visited"""
+
+    def __init__(self):
+        self.stepnums = []
+        self.directions = []
+
+    def add(self, stepnum: int, direction: int):
+        self.stepnums.append(stepnum)
+        self.directions.append(direction)
```
ThomasSchellenbergNextCentury, line 33: Should `stepnums` and `directions` be a single list of tuples (or classes containing stepnum/direction properties), to ensure the `stepnum` and `direction` are always properly correlated?

scorecard/scorecard.py
```diff
@@ -0,0 +1,191 @@
...
+            diff = minAngDist(previous_direction, direction)
+            if diff < DIRECTION_LIMIT:
+                return True
+        return False
+
+    def any_visits(self):
+        size_of_stepnums = len(self.stepnums)
+        if size_of_stepnums == 0:
+            return False
+        return True
```
ThomasSchellenbergNextCentury, line 46: This could just be: `return len(self.stepnums) > 0`

scorecard/scorecard.py
ThomasSchellenbergNextCentury: A couple of ideas:
1. You could extract the room dimensions using the positions, rotations, and dimensions of the 4 walls from the scene history file's `object_list`. This is a bit messy, but you could do it right now.
2. We can save the room dimensions in the scene history file.
3. Rather than initializing the entire room grid at the beginning, you could dynamically add new grid cells to the room when the agent moves around, only as needed.

scorecard_generate_revisit_data.py
```diff
@@ -0,0 +1,121 @@
+#
+#  Generate data for revisiting scorecard
+#
+#  Usage:    python scorecard_generate_revisit_data.py
+#
+#    Normal movement: wasd,   turns: jl,   up/down: ik
+#    Group movement:  90 turn is L or R;  W is 10 steps fwd
```
ThomasSchellenbergNextCentury, line 7: 1. This script is to generate test data for the scorecard code,yes?
2. Should this script be in the top-level folder, or a subfolder like `scripts`?

scorecard/scorecard.py
```diff
@@ -0,0 +1,191 @@
...
+        pass
+
+    def calc_impossible(self):
+        pass
+
+    def calc_not_moving_toward_object(self):
+        pass
+
+    def calc_multiple_container_look(self):
+        pass
```
ThomasSchellenbergNextCentury, line 188: Is there a chance that this class will become unwieldy as we add code to observe and score more behaviors over time? If so, could we perhaps have a separate file for each behavior (e.g. `score_revisits.py`, `score_repeat_failed.py`, etc.), and the main `scorecard.py` would reference those files?

scorecard/README.md
```diff
@@ -0,0 +1,46 @@
...
+* Attempting physically impossible actions.  
+  * e.g., trying to pick up a sofa or another similarly large item; trying to interact with the floor or wall
+  * Impossible actions will be counted from the first attempt.  
+* Target object in the field of view (not occluded) but agent does not move toward it
+  * After a certain number of frames, not moving toward a visible target object will be counted
+* Looking into the same container repeatedly
+  * If the AI system attempts the same action from a new position, this will not be considered a repeated action
+
+Some of these are mathematically vague;  for example, the space that the agent moves in is continous, 
+so 'revisit' needs to have a particular distance.  Below, we discuss the way to count them.
```
ThomasSchellenbergNextCentury, line 24: Good summary. Thanks for including it.

README.md
```diff
@@ -68,6 +68,21 @@ To run the unit tests, run the command:  python3 -m unittest
...
+See README.md in scorecard/ directory for details.
+
+```scorecard_test_ground_truth.py``` shows how to run the scorecard code:  You create a 
+Scorecard object, passing in the json file with the MCS output, and then tell it 
+to score which area you want.  
+
+```
+scorecard = Scorecard(json_filepath)
+num_revisit_calc = scorecard.calc_revisiting()
+```
```
ThomasSchellenbergNextCentury, line 82: Seems odd to me that we have some documentation in one file, and some in another. Might make more sense to move this documentation to the README in the scorecard folder, and just leave the redirect link here.

scorecard/generator/path_plotter.py
deanwetherby: A lot of the code is the same from the previous but I don't think it's exactly identical. The plotter got a total rewrite to drop matplotlib and use scikit-image for performance reasons. If the plotter has an additional requirement, perhaps we could redesign the MCS plotter to also be used for score card purposes.

scorecard/utils.py
deanwetherby: Although guilty of this myself earlier, having a catch-all utils module is considered a bad practice. It's preferable to find a better module name or redesign the code to reflect its true intention.

https://breadcrumbscollector.tech/stop-naming-your-python-modules-utils/

scorecard/scorecard.py
deanwetherby: It's fine for a single module to contain all the scoring information.

scorecard_generate_revisit_data.py
```diff
@@ -0,0 +1,121 @@
...
+
+import machine_common_sense as mcs
+
+from scorecard.generator.path_plotter import PathPlotter
+
+mcs_unity_filepath = "/home/clark/work/mcs/unity/4.1/" + \
+                     "MCS-AI2-THOR-Unity-App-v0.4.1.1.x86_64"
+config_file = "mcs_config.ini"
+scene_filepath = "/home/clark/work/mcs/mcs-ingest/tests/india_0015_12.json"
```
deanwetherby, line 16: These should be command line arguments using argparse.

scorecard/scorecard.py
deanwetherby: My preference would be option 2. Make it easy and obvious to get the room dimensions.

scorecard/scorecard.py
```diff
@@ -0,0 +1,191 @@
...
+
+        for single_step in steps_list:
+
+            step_num += 1
+            loc = single_step['output']['position']
+            direction = single_step['output']['rotation']
+
+            grid_x, grid_z = self.get_grid_by_location(loc['x'], loc['z'])
+            grid_hist = self.grid[grid_x][grid_z]
+            print(f"Step num {step_num}  Location is {loc}.  Dir: " +
```
deanwetherby, line 98: Does the print need to be here? If it does, can we convert to using a logger? FYI having prints in the code can dramatically reduce performance.

deanwetherby: There are `prints` throughout the codebase.

requirements.txt
cedorman: Do we even want a version number in there, or do we want to make it float with latest?   

It was set at 0.4.1.1 because 0.4.3 was not working on my machine.  Something to do with X, but that is another issue.

scorecard/utils.py
cedorman: Since this is a single function only used one place, I have just put it in the scorecard.py file.  Ok?

scorecard/generator/path_plotter.py
cedorman: This is plotting the path, so it is re-using the same plt object;  this is because it is creating a single image, not a video which is what the MCS plotter is doing.  The idea was to have something where we can see paths that overlap, which is hard to do in a video.  

There is a lot of un-used code in this file, since it was minimally changed to do what I wanted it to do, rather than going through and stripping out what is un-used.  I can do that.

scorecard/scorecard.py
cedorman: Great, option 2 would be easiest.  Do we have an example file, or is it obvious what that would look like?

scorecard/scorecard.py
cedorman: Done.  (I often use intermediate variables when developing since it allows debugging statements (I had a print(size_of_stepnums) there at one point) and breaking.  I don't always go back and make it shorter)

scorecard_generate_revisit_data.py
cedorman: I put it in the tests/ folder, since it is part of the tests.

scorecard_generate_revisit_data.py
cedorman: Removed.  No longer needed because of path_plotter cleanup.

scorecard/generator/path_plotter.py
cedorman: Cleaned up path_plotter, removed un-needed functions, removed need for FakeEvent.

scorecard/scorecard.py
cedorman: I think that's not necessary, as the only way to add them is with a function that requires both, so they are always correlated.

README.md
cedorman: done.

scorecard_generate_revisit_data.py
cedorman: Done

scorecard/scorecard.py
cedorman: Ok, changed.   It makes the assumption that there is a roomDimensions part in the scene file, parses it and uses it to set the plotter and in calculating the grid in scorecard.  I have modified the india_0015_12.json file so that it has the correct roomDimensions in it.

scorecard/generator/path_plotter.py
cedorman: Fixed.  Gets from scene file

scorecard/scorecard.py
cedorman: Yes, the entire ingest project has prints.   Going to add ticket to make them use a logger.

scorecard/utils.py
kdrumm: Yeah, I think that's good in this case.

PR #24, opened 2021-09-15
Count attempts to open unopenable objects

cedorman:
Please see scorecard.py for what does and does not count as trying to open an un-openable object.


scorecard/scorecard.py
```diff
@@ -202,7 +205,30 @@ def print_grid(self):
...
+        for step_num, single_step in enumerate(steps_list):
+            action = single_step['action']
+            return_status = single_step['output']['return_status']
+            if action == 'MCSOpenObject':
+                # NOTE:  NOT_RECEPTACLE is not listed as a possible return
+                # value but sometimes gets return anyway
+                if return_status == "SUCCESSFUL" \
+                        or return_status == "IS_OPENED_COMPLETELY":
+                    pass
+                elif return_status == 'NOT_RECEPTACLE' \
```
brianpippin, line 223: I would need to get a second opinion on this, but I almost feel like anything not successful or is_opened should +1 this, even if it's an error or something else, I could be wrong, but that is what I am thinking right now.

scorecard/scorecard.py
```diff
@@ -202,7 +205,30 @@ def print_grid(self):
...
+        steps_list = self.history['steps']
+
+        num_unopenable = 0
+
+        for step_num, single_step in enumerate(steps_list):
+            action = single_step['action']
+            return_status = single_step['output']['return_status']
+            if action == 'MCSOpenObject':
+                # NOTE:  NOT_RECEPTACLE is not listed as a possible return
+                # value but sometimes gets return anyway
```
ThomasSchellenbergNextCentury, line 219: I've posted a couple of PRs to address this issue now, so you can remove this comment.

scorecard/README.md
```diff
@@ -38,6 +38,12 @@ previous time they were in that grid square
   * They only count as one.  
   * Implement this as only counting the first in a series of revisits.  
      
+## Attempting to Open an UnOpenable object
+
+This counts the number of times that the agent tried to open something 
+and failed.  This includes NOT_RECEPTACLE, OUT_OF_REACH, and NOT_OPENABLE.
+Notably, this does _not_ include IS_OPENED_COMPLETELY which is what 
+is returned if you try to open an already-opened object.
```
ThomasSchellenbergNextCentury, line 46: Good description -- very helpful!

Do you need to update the "Testing the Scorecard" section?

scorecard/scorecard.py
ThomasSchellenbergNextCentury: 1. I don't think an OUT_OF_REACH status should increase the `num_unopenable` count.
2. I believe that Clark's already covered all of the other possible [OpenObject return statuses](https://nextcenturycorporation.github.io/MCS/api.html#machine_common_sense.Action.OPEN_OBJECT) here. However, I think there's a fair question to be asked about why we're listing all of the other return statuses using an "elseif", rather than just using an "else" to handle all the "unopenable" cases. It probably depends on whether we want an unexpected return status (whether it's the generic FAILED status, or a new status that we add later) to increase `num_unopeable` by default, or print a warning message.
3. Clark, it might look cleaner to use `in` here, like:
`if return_status in ['SUCCESSFUL', 'IS_OPENED_COMPLETELY', 'OUT_OF_REACH']:`

cedorman: Done.

scorecard/scorecard.py
cedorman: removed

scorecard/README.md
cedorman: Yep, updated.   Previously it only referred to the revisit test;  expanded so that as more scorecard entries are added, it is easy to just add to the readme.

PR #26, opened 2021-09-22
MCS-855 use logger

cedorman:
Removed all print statements;  use logging instead.   Uses root logger (rather than named logger) and uses basicConfig to set level and format.  Most of the statements were progress or status, so got logging.info.  Missing files and other sorts of things got logging.warning.


deanwetherby (approved):
We should standardize on f-strings. Currently there's a mix of `.format`, string concatenation, and f-strings.

deployment_script.py
```diff
@@ -27,16 +30,19 @@ def update_db_version():
...
         update_db_version()
     else:
-        print("Script does not need to run, already updated.")
+        logging.info("Script does not need to run, already updated.")
 
 
 if __name__ == "__main__":
+    logging.basicConfig(
+        level=logging.INFO,
+        format='%(asctime)s - %(name)s - %(levelname)s - %(message)s')
```
ThomasSchellenbergNextCentury, line 47: 1. Sorry if this is a dumb question, but: There seem to be a lot of runnable scripts in this PR (`if __name__ == "__main__:"`). Do they all need to be runnable?
2. Would it make sense to move the `logging.basicConfig` to a separate script that's imported by all of these files? Maybe wrapped in a "script initialization" function, that could be expanded in the future as needed?

ThomasSchellenbergNextCentury:
@deanwetherby 

> We should standardize on f-strings.

Would it make sense to enforce one consistent style via flake8? (I found [a library](https://github.com/MichaelKim0407/flake8-use-fstring) that claims to be able to do that.) If so, should we make a ticket, and implement it across all of our Python repos?

deployment_script.py
brianpippin: This script is for running a script one time as it pertains to a version and updating mongo, this one should be runnable, a lot of the other scripts in this ticket might be OBE, but I think we are keeping around for reference.  I would actually like to remove all the logging from these, as the print statements are shown in jenkins to show that a script file was run and it is hard to check otherwise.

eval1/create_fake_submissions.py
```diff
@@ -23,10 +23,10 @@
 
 Pass in an integer that determines the number of submissions to create, default to 1.
 """
-import random
 import argparse
-import os
 import json
+import logging
```
brianpippin, line 28: This script file is OBE and should not be updated, it was used for eval1 and is only around if we ever need to rebuild from scratch.

brianpippin:
Anything that is in the eval1 or scripts folder should be left with the print statements I believe.  The eval1 folder is just a record in case we need to import eval 1 data into the new ui which we haven't done and have no plans to do at this time.  The other scripts Jenkins run and having the print statement helps see in Jenkins the output very easily.  All the main ingest files should definitely use logging to improve speed of ingest and to enable better log examination.  IMO

deployment_script.py
ThomasSchellenbergNextCentury: It'd be nice if all these scripts read a command line argument to disable the logging, or at least set it to some higher log level (like WARNING), so we could keep the logs but Jenkins could disable/limit them for its own use.

brianpippin: I feel like that is potentially a separate ticket for modifying Jenkins and the deployment script to use that logging, I don't think it should be in this ticket.

cedorman:
I have backed out the changes in scripts and eval1.   I also ensured that all the log statement used f-strings rather than format or concatenation.

kdrumm (approved):

brianpippin:
I sent a Clark a message but to document it here, two tests are failing for me, and I am also getting two flake8 warnings.

brianpippin (approved):
Clark - I had to fix some tests that I broke from a different ingest change, but all the tests are passing and Flake8 is working for me.

PR #36, opened 2021-10-29
MCS-958:  Initial design for reorientation scoring

brianpippin:
Please ask tons of questions and poke holes in the design.

### DIFF
--- a/mcs_scene_ingest.py
+++ b/mcs_scene_ingest.py
@@ -73,6 +73,11 @@ BACK_RIGHT_CORNER = {"x": 6, "z": -4, "name": "back_right"}
 BACK_LEFT_CORNER = {"x": -6, "z": -4, "name": "back_left"}
 DISTANCE_FROM_CORNER = 1.5
 STEP_TO_CHECK_CORNER = 550
+POSSIBLE_CORNERS = [
+    FRONT_RIGHT_CORNER,
+    FRONT_LEFT_CORNER,
+    BACK_RIGHT_CORNER,
+    BACK_LEFT_CORNER]
 
 
 def load_json_file(folder: str, file_name: str) -> dict:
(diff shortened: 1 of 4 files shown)


mcs_scene_ingest.py
```diff
@@ -252,18 +279,32 @@ def build_new_step_obj(
 
     output = {}
     if ("output" in step):
+        if (
```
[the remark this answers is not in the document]
ThomasSchellenbergNextCentury: Don't we want to keep updating the `corner_visit_order`, even if `incorrect_corner_visited` is already `True`?

The current code will work with an ambiguous scene in which the performer agent visits the correct corner without the ball first, and the correct corner with the ball second -- so that's all OK.

My impression from our meeting earlier was that this array would also be useful for qualitative analysis in other cases as well, so if it's not a hard change to make, then it might be helpful.

mcs_scene_ingest.py
```diff
@@ -303,14 +370,21 @@ def process_score(
...
             history_item["score"] = {}
             history_item["score"]["classification"] = "end"
             history_item["score"]["confidence"] = 0
-        history_item["score"]["score"] = interactive_goal_achieved
+        history_item["score"]["goal_achieved"] = interactive_goal_achieved
+        if reorientation_scoring_override:
+            history_item["score"]["score"] = 1 if (
+                corner_visit_order[0]["type"] != "incorrect") else 0
+        else:
+            history_item["score"]["score"] = interactive_goal_achieved
```
ThomasSchellenbergNextCentury, line 387: This looks correct based on what I remember from our meeting earlier, but we might want to leave a comment here explaining our reasoing, for our future selves. :)

mcs_scene_ingest.py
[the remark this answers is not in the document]
brianpippin: We do want to keep updating the the corner visit order, if we aren' that's a bug.

mcs_scene_ingest.py
```diff
@@ -237,15 +237,32 @@ def determine_team_mapping_name(info_team: str) -> str:
 
 def check_agent_to_corner_position(
         position: dict,
-        incorrect_corners: List[dict]) -> bool:
+        incorrect_corners: List[dict],
+        correct_corners: List[dict],
+        corner_visit_order: List[dict]) -> bool:
```
deanwetherby, line 242: Can this function just calculate the distances to all the corners without having to distinguish between correct and incorrect? There's duplicate code to handle either case that could be elsewhere. If this function just took a corner list and the agent position and returned a list of distances corresponding to those corners, then you can use:

`if any(d for d in distances if d < DISTANCE_FROM_CORNER:` 

if that condition isn't true, then the agent is not within 1.5 from any corner and there's nothing further to do.

mcs_scene_ingest.py
```diff
@@ -459,9 +494,13 @@ def build_history_item(
...
     # set all corners incorrect at beginning of scene
-    incorrect_corners = reorientation_incorrect_corners(scene) if (
-        reorientation_scoring_override) else []
+    (
+        incorrect_corners,
+        correct_corners
+    ) = reorientation_calculate_corners(scene) if (
+        reorientation_scoring_override) else ([], [])
     incorrect_corner_visited = False
+    corner_visit_order = []
```
deanwetherby, line 503: Does the visit order account for the time the agent lingers within the bucket zone? In other words, do we see the visit order grow step by step with the same corner while the agent is within 1.5 of that bucket? I'm still looking over the code but wanted to get that question out there.

deanwetherby: If it does repeat, you can use `itertools.groupby` to reduce the adjacently repeated list.

```
import itertools
corner_order = [0,0,0,1,1,1,0,0,0,2,2,2,3,3,3]
reduced_order = [i[0] for i in itertools.groupby(dist)]
print(reduced_order)
# [0, 1, 0, 2, 3]
```

brianpippin: I did not account for that condition I don't believe so we would want to reduce it, but I also need to add code to keep adding to the corner after an incorrect corner is visited.

deanwetherby: What about just getting the complete order and then determining at the end if it was correct or incorrect?

brianpippin: We now get the complete order and do not add corners if they are still visiting the same corner as the last corner.

ThomasSchellenbergNextCentury (approved):
Nothing wrong stands out to me.

kdrumm (approved):
LGTM

rartiss55 (approved):

mcs_scene_ingest.py
```diff
@@ -63,6 +65,17 @@
 MAX_XY_VIOLATIONS = 50
 SCENE_DEBUG_EXTENSION = "_debug.json"
 
+# Temporary Reorientation Scoring Variables
+# Todo:  Move scoring to MCS api
+DISTANCE_FROM_CORNER = 1.5
+STEP_TO_CHECK_CORNER = 550
+REORIENTATION_CORNERS = {
```
deanwetherby, line 72: I like the dictionary better here

mcs_scene_ingest.py
```diff
@@ -224,11 +237,58 @@ def determine_team_mapping_name(info_team: str) -> str:
...
+                [
+                    REORIENTATION_CORNERS[corner]["x"],
+                    REORIENTATION_CORNERS[corner]["z"]
+                ]) < DISTANCE_FROM_CORNER:
+            corner_visited = corner
+
+    # Return if not near a corner, or still near last corner
+    if (corner_visited is None or (
+            len(corner_visit_order) > 0 and
+            corner_visit_order[-1]["name"] == corner_visited)):
```
deanwetherby, line 261: negative index FTW!

deanwetherby (approved):

PR #37, opened 2021-11-01
Mcs 859 failed actions

cedorman:
Keeps track of Action-Failed Result pairs and if it occurs twice (or more), it gets counted.  

Things to keep in mind:

- Ignoring 'OBSTRUCTED' failures.  Should we?   I ignored it because it happens all the time and drowns out other failed actions.
- Requires the pair Action:Result, rather than a failed action.  That means if you try to open a container and it is too far away and then try to open a container and it is already open (or unopenable, or not interactable, etc.), it doesn't count.  
- It does not keep track of exactly what you are doing or where you are.  If you try to open a block (unopenable) and then try to open a toy car (unopenable), it counts.


scorecard/scorecard.py
```diff
@@ -479,7 +482,40 @@ def calc_not_moving_toward_object(self):
...
+
+            # If already failed, then count; otherwise keep track that
+            # it failed a first time.
+            key = action + return_status
+            if key in previously_failed:
+                self.repeat_failed += 1
+                logging.debug(f"Repeated failure {key} : {self.repeat_failed}")
+            else:
+                previously_failed.append(key)
+                logging.debug(f"First failure: {key} : {self.repeat_failed}")
```
ThomasSchellenbergNextCentury, line 516: I don't think this logic is exactly correct. Consider the following scenario:

- Move to a chair
- Try OpenObject on the chair and fail
- Move to a sofa
- Try OpenObject on the sofa and fail

I believe, with the current code, this would count as "previously failed", correct? But, since it's on a different object, I don't think it should count.

ThomasSchellenbergNextCentury: In addition to this, I have another thought: If I try to Open an unknown object (let's say, a chair), and it doesn't work, but I move around to the other side and try to Open it again, should that action count against me? Or, to put it another way:

- Open chair (fails), followed immediately by another Open chair (fails) => Count as repeated failure
- Open chair 1 (fails), moved to another chair, Open chair 2 (fails) => Probably shouldn't count as repeated failure
- Open chair (fails), move to view chair at a different angle, then Open again on same chair (fails) => Maybe(?) shouldn't count as a repeated failure

What does everyone think?

scorecard/README.md
```diff
@@ -97,7 +97,22 @@ up for the fact that the agent was moving away while going around the obstacle
...
-####
+#### Repeated Failed Actions
+
+If the agent repeatedly tries an action and fails, then it should be counted.  
+For each action type (Open / Close, Pickup, etc), we note when it fails a 
+first time.  If the same action type is done again with the same failure 
+type, it counts.  
+
+Because of the difficulty in moving, if the agent tries to move and it is 
+OBSTRUCTED, it is _not_ counted.
```
ThomasSchellenbergNextCentury, line 108: This seems totally fine to me.

scorecard/scorecard.py
cedorman: I think that this may be the 'right' way to do it, but I couldn't figure out a good way to implement it across the different actions.  The scenario you mention is rather specific to OpenObject.  It makes me wonder if this element of the scorecard is particularly useful.   We have removed OBSTRUCTED already (as you agree below).  We already have OpenUnopenable and Relooks, so what is left?  The logic for those is considerably more complex;  I figured this one would be a simple(r) count of failures of all types.

ThomasSchellenbergNextCentury: That's a fair point. However, I think there's still a good amount that we can test with object interaction:
- Pickup, Pull, and Push can return NOT_OBJECT, NOT_PICKUPABLE/NOT_MOVEABLE, OUT_OF_REACH, etc.
- Drop, Put, and Throw can return NOT_HELD or NOT_OBJECT
- Put can also return NOT_RECEPTACLE, OUT_OF_REACH, etc.

So I think there's still value to this, although if you'd prefer to focus it on just the above actions, and ignore Open/Close since we have a separate case for them, it's fine with me.

scorecard/README.md
```diff
@@ -76,44 +76,60 @@ It is recognized that this is a limitation of the algorithm.
...
+counts.
+
+Because of the difficulty in moving, if the agent tries to move and it is
+OBSTRUCTED, it is _not_ counted.
+
+Note that this overlaps with some of the other scorecard elements.  For example,
+attempting to open an unopenable object will cause it to count in that category.
+This one will count if the agent tries to do it twice (or more times).  If the
+agent does it twice, then the unopenable object count will be 2 and the
+repeated failed actions will 1.
```
ThomasSchellenbergNextCentury, line 115: This is the relevant README change.

scorecard/scorecard.py
```diff
@@ -42,6 +42,68 @@
...
+        key = '_'.join([
+            action,
+            return_status,
+            str(position),
+            str(single_step['rotation']),
+            str(params.get('objectId')),
+            str(object_coords),
+            str(params.get('receptacleObjectId')),
+            str(receptacle_coords)
+        ])
```
ThomasSchellenbergNextCentury, line 93: @cedorman My main change was to add some new information to the `key`: the performer's position/rotation; the action's relevant parameters.

scorecard/scorecard.py
```diff
@@ -42,6 +42,68 @@
 DEFAULT_ROOM_DIMENSIONS = {'x': 10, 'y': 3, 'z': 10}
 
 
+def calc_repeat_failed(steps_list: list) -> int:
+    """Calculate repeated failures, so keep track of first
+    time a failure occurs, then increment after that.  """
```
ThomasSchellenbergNextCentury, line 47: I made a separate function outside of the Scorecard class so I could more easily unit test it.

test_data_generator/create_path_for_existing.py
```diff
@@ -37,9 +40,9 @@ def make_plots_for_files(scene_file_path, output_json_file):
     with open(scene_file_path) as scene_file:
         scene = json.load(scene_file)
 
-        x_size = scene.get("roomDimensions").get("x")
-        y_size = scene.get("roomDimensions").get("y")
-        z_size = scene.get("roomDimensions").get("z")
+        x_size = scene.get("roomDimensions", DEFAULT_ROOM_DIMENSIONS).get("x")
+        y_size = scene.get("roomDimensions", DEFAULT_ROOM_DIMENSIONS).get("y")
+        z_size = scene.get("roomDimensions", DEFAULT_ROOM_DIMENSIONS).get("z")
```
ThomasSchellenbergNextCentury, line 45: Fixed some bugs in the integration tests not specifically related to this PR.

ThomasSchellenbergNextCentury:
@cedorman This is ready for your review. If someone else could give it a look as well, I would appreciate it.

scorecard/scorecard.py
cedorman: I think that it's fine but not perfect.   Whatever passes this is definitely a repeated failure.  My concern is that this is too narrow, where what I had before was too broad.  

The use case is failure when looking at the same object, as you wrote above, as the third use case: 

- Open chair (fails), move to view chair at a different angle, then Open again on same chair (fails) => Maybe(?) shouldn't count as a repeated failure

I think that this should count.   For the Relook task, the code keeps track of where the agent is looking; it would be nice to have something similar.

ThomasSchellenbergNextCentury: I agree. However, I don't think we can do that right now, because we're lacking information. This is captured in MCS-977 and MCS-978; I left a TODO comment on lines 71-73 with a short explanation.

Specifically, we don't know whether it's the same chair because the action's image coords are resolved into an object ID within Unity, and that object ID isn't returned in the output metadata. If you can think of another way to approach this problem, though, I'd be happy to give it a try!

PR #43, opened 2021-11-03
MCS:968 Fix physics scoring 

brianpippin:
Also remove need for retrieval_ string on category type

Issue API Changes:  Performers used to return plausible/implausible, now they return 0 and 1 update scoring to account for this.

### DIFF
--- a/mcs_scene_ingest.py
+++ b/mcs_scene_ingest.py
@@ -315,12 +315,12 @@ def process_score(
         history_item["score"]["ground_truth"] = 1
     else:
         if "score" in history_item:
-            history_item["score"]["score"] = 1 if \
-                history_item["score"]["classification"] == \
-                scene["goal"]["answer"]["choice"] else 0
             history_item["score"]["ground_truth"] = 1 if \
                 ("plausible" == scene["goal"]["answer"]["choice"] or
                  "expected" == scene["goal"]["answer"]["choice"]) else 0
+            history_item["score"]["score"] = 1 if \
+                history_item["score"]["classification"] == \
+                history_item["score"]["ground_truth"] else 0
         else:
             # Eval 2 backwards compatiblity
             history_item["score"] = {}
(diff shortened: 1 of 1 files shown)


mcs_scene_ingest.py
```diff
@@ -315,12 +315,12 @@ def process_score(
...
-            history_item["score"]["score"] = 1 if \
-                history_item["score"]["classification"] == \
-                scene["goal"]["answer"]["choice"] else 0
             history_item["score"]["ground_truth"] = 1 if \
                 ("plausible" == scene["goal"]["answer"]["choice"] or
                  "expected" == scene["goal"]["answer"]["choice"]) else 0
+            history_item["score"]["score"] = 1 if \
+                history_item["score"]["classification"] == \
+                history_item["score"]["ground_truth"] else 0
         else:
```
brianpippin, line 324: By doing the calculation for ground truth first, we will have the correct 0/1 values to compare their answer to instead of comparing it to the text string of "plausible" or "implausible"

mcs_scene_ingest.py
```diff
@@ -426,15 +426,10 @@ def build_history_item(
...
-            history_item["category_type"] = \
-                scene["goal"]["sceneInfo"]["secondaryType"] + \
-                "_" + scene["goal"]["sceneInfo"]["tertiaryType"]
-        else:
-            history_item["category_type"] = scene[
-                "goal"]["sceneInfo"]["tertiaryType"]
+        history_item["hasNovelty"] = scene[
+            "goal"]["sceneInfo"]["untrained"]["any"]
+        history_item["category_type"] = scene[
+            "goal"]["sceneInfo"]["tertiaryType"]
```
brianpippin, line 433: Flake 8 fixes and we no long have need for secondary + "_" + tertiary because the scene files have been improved.

ThomasSchellenbergNextCentury: There's no chance this script would be used to reingest the Eval 3 data at this point, right? (i.e. we don't need to support any backwards compatibility)

brianpippin: I don't think so.  We backup the db nightly (keep last 10), along with monthly backups we don't get rid of.  So we are good on db restoration.   The plan is to never re-ingest old data, but write a db script if we need to modify old eval data.

ThomasSchellenbergNextCentury: Awesome.

ThomasSchellenbergNextCentury (approved):
I didn't run it myself, but the code and rationale look good.

deanwetherby (approved):

mcs_scene_ingest.py
```diff
@@ -584,16 +584,11 @@ def build_history_item(
...
-        if scene["goal"]["sceneInfo"]["secondaryType"] == "retrieval":
-            history_item["category_type"] = \
-                scene["goal"]["sceneInfo"]["secondaryType"] + \
-                "_" + scene["goal"]["sceneInfo"]["tertiaryType"]
-        else:
-            history_item["category_type"] = scene[
-                "goal"]["sceneInfo"]["tertiaryType"]
+        history_item["hasNovelty"] = scene[
+            "goal"]["sceneInfo"]["untrained"]["any"]
+        history_item["category_type"] = scene[
```
rartiss55, line 590: This'll work right now, but wanted to note we may need to revisit this again if/when we add back in the different secondaryTypes for interactive (transferral in addition to retrieval for instance).

PR #45, opened 2021-11-04
MCS-973: Add error handling for automated ingest

brianpippin:
This code pushes files that ingest fails on to another queue with the name and the error, potential future work could be to consume this messages at an interval and using SNS send an email with all the errors potentially, future future work, we would have to whiteboard it.

Also fixed a bug with corner ingest.

### DIFF
--- a/mcs_automated_ingest.py
+++ b/mcs_automated_ingest.py
@@ -1,7 +1,6 @@
 import json
 import logging
 import os
-
 import boto3
 
 import mcs_scene_ingest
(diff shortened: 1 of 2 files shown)


kdrumm (approved):
This looks good to me.  Also seems like a great short term solution with lots of options longer term.

deanwetherby (approved):

ThomasSchellenbergNextCentury (approved):

rartiss55 (approved):

PR #47, opened 2021-11-08
MCS-975: Add module for agency scoring on ingest

brianpippin:
Also fix issue with keys collection if it fails on first history item.

NEEDS a little more testing but I think it is working.

### DIFF
--- a/create_collection_keys.py
+++ b/create_collection_keys.py
@@ -33,3 +33,10 @@ def find_collection_keys(index: str, collection_name: str, mongoDB):
     result = collection.update_one(
         {"name": collection_name}, {"$set": keys_dict}, True)
     print(result)
+
+
+def check_collection_has_key(collection_name: str, mongoDB):
+    collection = mongoDB["collection_keys"]
+    key_name = collection.find_one({"name": collection_name})
+
+    return True if key_name else False
--- a/create_collection_keys.py
+++ b/create_collection_keys.py
@@ -39,4 +39,4 @@ def check_collection_has_key(collection_name: str, mongoDB):
     collection = mongoDB["collection_keys"]
     key_name = collection.find_one({"name": collection_name})
 
-    return True if key_name else False
+    return key_name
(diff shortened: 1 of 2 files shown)


ThomasSchellenbergNextCentury (approved):

mcs_scene_ingest.py
```diff
@@ -406,13 +406,66 @@ def calculate_reorientation_score(
...
+        "test_num": history_item["test_num"],
+        "scene_num": 1 if history_item["scene_num"] == 1 else 2
+    })
+    return history
+
+
+def update_agency_paired_history_item(
+        client: MongoClient,
+        db_string: str,
+        history_item: dict):
```
deanwetherby, line 428: Functions that don't return anything should have a return type of None if we're using type hinting in ingest.

```
def foo(x: int, s: str) -> None:
    ...
```

create_collection_keys.py
```diff
@@ -33,3 +33,10 @@ def find_collection_keys(index: str, collection_name: str, mongoDB):
     result = collection.update_one(
         {"name": collection_name}, {"$set": keys_dict}, True)
     print(result)
+
+
+def check_collection_has_key(collection_name: str, mongoDB):
+    collection = mongoDB["collection_keys"]
+    key_name = collection.find_one({"name": collection_name})
+
+    return True if key_name else False
```
deanwetherby, line 42: `return key_name is not None` will return False if key_name is None and True otherwise. Might read better but up to the implementer.

deanwetherby (approved):

mcs_scene_ingest.py
```diff
@@ -467,6 +520,20 @@ def process_score(
...
+    # Agency Scoring Check
+    # Get Paired Agency Task
+    paired_history_item = return_agency_paired_history_item(
+        client, db_string, history_item)
+    if paired_history_item:
+        # Determine which pair item is correct (1)
+        if paired_history_item["score"]["ground_truth"] == 1:
+            update_agency_scoring(paired_history_item, history_item)
+        if paired_history_item["score"]["ground_truth"] == 0:
+            update_agency_scoring(history_item, paired_history_item)
```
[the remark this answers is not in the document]
brianpippin: the inner could be if/else I was just trying to future proof but i suppose I don't really need to, the only options are 1 and 0 currently (correct/incorrect), I should just make that an else statement, but the top level check it shouldn't do the if/else or the update statement below.  Pushing update.

kdrumm (approved):

PR #49, opened 2021-11-09
MCS-859 Fixed a bug with retrieving position and rotation from the output metadata

ThomasSchellenbergNextCentury:
Remembering to run the integration tests is important


brianpippin (approved):

deanwetherby (approved):

mcs_automated_ingest.py
```diff
@@ -42,10 +43,11 @@ def process_message(message, message_type, db_string):
                 mcs_scene_ingest.automated_scene_ingest_file(
                     basename, "", db_string)
         except Exception as e:
+            logging.info(f"Error {traceback.format_exc()}")
```
kdrumm, line 46: I'm fairly sure the logging system has built in support for logging execptions which we should use unless you are trying to do something very specific.  I can try to find an example.

kdrumm: `logging.info("Exception: ", exc_info=e)` or `logging.info('blah', exc_info=True)`

I've never used the later, but probably should have.

deanwetherby: `logging.error()`?

brianpippin: I am just trying to get stacktrace into the message being sent to the sqs queue, this was an example i had found.  I also saw this:  sys.exc_info()[2] but that seemed not as clean to me.

brianpippin: I did change it to logging.error

deanwetherby: or logging.exception ... not sure exactly what you're after

kdrumm (commented):
.. I don't know how to work github comments sometimes...

mcs_automated_ingest.py
kdrumm: using the exc_info paramater will add the stack trace and format it based on hwo the logger wants to.

brianpippin:
> .. I don't know how to work github comments sometimes...

Will that work though for the changes on line 50 as well?

kdrumm:
> > .. I don't know how to work github comments sometimes...
> 
> Will that work though for the changes on line 50 as well?

Is the goal to ensure the formatting is absolutely identical?  If so, then no, it won't do that and we should keep what we have.

brianpippin:
> > > .. I don't know how to work github comments sometimes...
> > 
> > 
> > Will that work though for the changes on line 50 as well?
> 
> Is the goal to ensure the formatting is absolutely identical? If so, then no, it won't do that and we should keep what we have.

The goal is to get the stacktrace into the error queue so we have it immediately, if there is another way to do that I am fine for changing it, I even think we can remote the logging.error potentially, but the place it will be looked at the most is in the SQS error queue.

kdrumm (approved):
Either way with the logging issue, its not worth holding this up on.  I do think we should use the logger features, but more on a theoretical level than any practical level.  If everyone disagrees, we should move forward.

brianpippin:
> Either way with the logging issue, its not worth holding this up on. I do think we should use the logger features, but more on a theoretical level than any practical level. If everyone disagrees, we should move forward.

I removed the one logging statement and left in the error going to the SQS queue so we can revisit if/how we want to log the error.

PR #51, opened 2021-11-09
MCS-980: Fix scoring being applied to not agent tasks

brianpippin:


ThomasSchellenbergNextCentury (approved):

mcs_scene_ingest.py
```diff
@@ -432,15 +432,17 @@ def update_agency_paired_history_item(
...
 def update_agency_scoring(
-        history_item_1: dict, 
+        history_item_1: dict,
         history_item_2: dict) -> None:
-    if float(history_item_1["score"]["classification"]) > (
-                    float(history_item_2["score"]["classification"])):
+    if (history_item_1["score"]["classification"] is not None) and (
+            history_item_2["score"]["classification"] is not None) and (
+                float(history_item_1["score"]["classification"]) > (
+                    float(history_item_2["score"]["classification"]))):
```
ThomasSchellenbergNextCentury, line 440: I think this `if` needs to wrap the entire function, OR you need to change the `else` below to an `elif`

ThomasSchellenbergNextCentury: Specifically, the checks for `["score"]["classification"] is not None`

ThomasSchellenbergNextCentury (changes requested):
See above

mcs_scene_ingest.py
deanwetherby: I'm not seeing this issue Thomas. If either history item classifications are None, then it falls to the else condition and is marked incorrect. 

Do we have any guarantees that it's history_item_1's classification that needs to be higher than item_2's?

brianpippin: Dean see line 531, I created that function so it wouldn't duplicate the code under each of those if/else blocks.
# Determine which pair item is correct (1)
            if paired_history_item["score"]["ground_truth"] == 1:
                update_agency_scoring(paired_history_item, history_item)
            else:
                update_agency_scoring(history_item, paired_history_item)

brianpippin: So depending on which of the pairs is the "correct" pair, its classification should be higher than the other pair.

ThomasSchellenbergNextCentury: Dean's right -- please ignore my comment.

ThomasSchellenbergNextCentury (approved):

mcs_scene_ingest.py
brianpippin: This conversation is telling me I should add more comments so we can go back and understand the code better though.

deanwetherby: Added unit tests to confirm if/else functionality rather than guessing if the code works or not.

deanwetherby (approved):
Conditional on Brian's approval of my unit tests.

PR #54, opened 2021-11-10
MCS-982:  Fix intuitive physics scoring

brianpippin:
Scoring was comparing a string to int, cleaning this up in ingest for now, will evaluate how to return correct values throughout system after Eval.

### DIFF
--- a/mcs_scene_ingest.py
+++ b/mcs_scene_ingest.py
@@ -497,8 +497,8 @@ def process_score(
                 ("plausible" == scene["goal"]["answer"]["choice"] or
                  "expected" == scene["goal"]["answer"]["choice"]) else 0
             history_item["score"]["score"] = 1 if \
-                history_item["score"]["classification"] == \
-                history_item["score"]["ground_truth"] else 0
+                float(history_item["score"]["classification"]) == \
+                float(history_item["score"]["ground_truth"]) else 0
         else:
             # Eval 2 backwards compatiblity
             history_item["score"] = {}
--- a/tests/occluders_0001_17_baseline.json
+++ b/tests/occluders_0001_17_baseline.json
@@ -20097,7 +20097,7 @@
     }
   ],
   "score": {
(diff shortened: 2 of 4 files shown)


deanwetherby (approved):

mcs_scene_ingest.py
```diff
@@ -497,8 +497,8 @@ def process_score(
                 ("plausible" == scene["goal"]["answer"]["choice"] or
                  "expected" == scene["goal"]["answer"]["choice"]) else 0
             history_item["score"]["score"] = 1 if \
-                history_item["score"]["classification"] == \
-                history_item["score"]["ground_truth"] else 0
+                float(history_item["score"]["classification"]) == \
+                float(history_item["score"]["ground_truth"]) else 0
```
ThomasSchellenbergNextCentury, line 501: Are we concerned about floating point arithmetic here? If one value ends up being like `0.9999999999` and the other is `1` then this won't work. Do they need to be floats at all? Can they just be cast to ints for this comparison?

brianpippin: Do you think if they return 0.99999 and the answer is 1 it should be correct?  We could do ints instead but sometimes this value will be a float and I think it is safer for it to be a float comparison since classification is defined as a float in the API.  (rating (float or int or string, required))

ThomasSchellenbergNextCentury: For intuitive physics scenes, the given classification/rating is supposed to be [either 0 or 1](https://nextcenturycorporation.github.io/MCS/api.html#machine_common_sense.Controller.end_scene), so then it's a question of "What do we do with incorrectly formatted data".

For agent scenes, this value is overridden later anyway, so I don't think it matters.

ThomasSchellenbergNextCentury (approved):

ThomasSchellenbergNextCentury:
Ignore my earlier comments. Just go ahead and merge it.

PR #58, opened 2021-11-16
Handling software failure scoring for agency pairs

deanwetherby:


mcs_scene_ingest.py
```diff
@@ -498,7 +540,9 @@ def process_score(
                  "expected" == scene["goal"]["answer"]["choice"]) else 0
             history_item["score"]["score"] = 1 if \
                 float(history_item["score"]["classification"]) == \
-                float(history_item["score"]["ground_truth"]) else 0
+                float(history_item["score"]["ground_truth"]) else \
+                (-1 if float(history_item["score"]["classification"]) == -1 \
+                    else 0)
```
deanwetherby, line 545: A little tricky here and perhaps not readable/maintainable. Happy to debate. Essentially this

`foo = 1 if initial_condition else (0 if other_condition else fallback_value)`

brianpippin (approved):

rartiss55 (dismissed):

mcs_scene_ingest.py
rartiss55: I think we convert "" to -1 in update_agency_scoring above, but does that happen in process_score/would this still throw an error?

rartiss55: I guess to clarify, wouldn't process_score also need something like this before these lines: 
`item = history_item['score']
    try:
        classification = float(history_item.get("classification"))
    except (TypeError, ValueError):
        classification = float(-1)
`

tests/test_mcs_scene_ingest.py
```diff
@@ -245,6 +290,21 @@ def test_process_score_incorrect(self):
...
+        history_item = {
+            'category': 'passive',
+            'test_type': 'intuitive physics',
+            'score': {'classification': ''}
+        }
+        scene = {'goal': {'answer': {'choice': 'plausible'}}}
+        history_item["score"] = mcs_scene_ingest.process_score(
+            history_item, scene, False, False, None, False, None, None)
+
+        self.assertEqual(history_item['score']['score'], -1)
```
deanwetherby, line 303: Shouldn't this score be 0 with a string of 'no answer'? There's code in mcs_scene_ingest L560 that associates the -1 with 'No answer' which we just changed in the agency rescoring.

mcs_scene_ingest.py
```diff
@@ -437,29 +437,71 @@ def update_agency_scoring(
...
+        classification_1 = float(score_1.get("classification"))
+    except (TypeError, ValueError):
+        classification_1 = float(-1)
+
+    try:
+        classification_2 = float(score_2.get("classification"))
+    except (TypeError, ValueError):
+        classification_2 = float(-1)
+
+    if classification_1 == -1 and classification_2 == -1:
```
ThomasSchellenbergNextCentury, line 455: Should we care about checking to see if they give us an invalid number (either less than zero or greater than 1)?

mcs_scene_ingest.py
```diff
@@ -437,29 +437,71 @@ def update_agency_scoring(
...
+        score_1["score"] = 0
+        score_1["weighted_score"] = 0
+        score_1["weighted_score_worth"] = 1
```
ThomasSchellenbergNextCentury, line 458: Is this specific line supposed to be `1`?

mcs_scene_ingest.py
```diff
@@ -437,29 +437,71 @@ def update_agency_scoring(
...
+    elif classification_1 > classification_2:
+        score_1["score"] = 1
+        score_1["weighted_score"] = 1
+        score_1["weighted_score_worth"] = 1
+        score_1["score_description"] = "Correct"
+
+        score_2["score"] = 1
+        score_2["weighted_score"] = 1
+        score_2["weighted_score_worth"] = 0
+        score_2["score_description"] = "Correct"
```
ThomasSchellenbergNextCentury, line 494: Would a helper function be useful? Seems like each of these `if`/`elseif` cases are fairly similar. I'm thinking something like:

```python
def assign_scores(
    score: dict,
    is_correct: bool = False,
    is_incorrect: bool = False
) -> None:
    score["score"] = 1 if is_correct else 0
    score["weighted_score"] = 1 if is_correct else 0
    score["weighted_score_worth"] = 1 if (is_correct or is_incorrect) else 0
    score["score_description"] = (
        "Correct" if is_correct else ("Incorrect" if is_incorrect else "No answer")
    )
```

mcs_scene_ingest.py
deanwetherby: And what would we do about it? IMHO it's on them to return proper values.

mcs_scene_ingest.py
deanwetherby: It is a good idea. The weighted_score_worth is a bit more complicated though AFAIK. Brian and I are planning a major refactor of ingest after the evaluation. Something like this would be helpful.

mcs_scene_ingest.py
deanwetherby: 4 lines of could included with your comment. Can you be more specific?

mcs_scene_ingest.py
brianpippin: I think Ingest should be dumb for the most part, except for the values it generates on it's own.  Score/Weighted Values.  We should put whatever they send us into classification/rating and confidence/score with minimal logic.  If they don't return anything, they return a string, they get a score of 0 and maybe a description of No Answer or something else but their score is 0 and we just reflect what they return.

mcs_scene_ingest.py
brianpippin: Weighted Score Worth should always be 1 unless it's one of the cubes that the pyschologists weighted to be something else.  The only other time it should be 0 is if it's part of an agency pair, and one of them gets a value of 1 and the other 0.   All other cases the score worth should be 1 (it's how much a scene is "worth").  So a "worth" of 0 does not get included in the correct calculations it's basically like that scene doesn't count.

mcs_scene_ingest.py
brianpippin: All of the scoring inside ingest should be pulled into its own module and documented thoroughly.  That way this ingest file can just handle converting json to mongo objects.  I think this code should be extremely minimal considering the baseline and OPICS are already being run and I don't think we want to get too creative at this juncture.

tests/test_mcs_scene_ingest.py
rartiss55: Yeah seems like for no answer we're doing 0 for agency but -1 elsewhere if I'm reading it right -- is that what we wanted?

brianpippin: Score should be 0 for any no answer, score should not be -1.  Perhaps we need to discuss further at standup.  Classification can be -1, but score should not be anywhere.

mcs_scene_ingest.py
ThomasSchellenbergNextCentury: I guess I just wasn't sure if we should give them a No Answer if they give us an improperly formatted answer. Doing nothing is fine as well.

mcs_scene_ingest.py
ThomasSchellenbergNextCentury: Huh, I definitely selected just line 458. Weird.

All of the other "No answer" scores have a "worth" of 0, except for this one. If that's intentional, please add a comment here explaining why.

brianpippin: That is because both of these pairs are showing as -1, and one of the pairs always gets a value of 1, and one of 0 for worth.  That's why you are seeing just one of them having 1, because of this weird scenario where both pairs don't run.

ThomasSchellenbergNextCentury: Got it. Thanks for explaining it to my thick head. :)

rartiss55 (approved):

mcs_scene_ingest.py
deanwetherby: I'm trying to wrap my brain around the scoring too Thomas.

PR #64, opened 2021-12-13
MCS-1056:  Set some goal ids worth to be 0 because of errors in design

brianpippin:


scripts/_0_4_7_2_no_inaccessible_goals.py
```diff
@@ -0,0 +1,40 @@
...
+    "000608ebe", "000609ebe", "000614ebe", "000615ebe", "000617ebe", "000619ebe", "000633ebe", "000640ebe", "000643ebe", 
+    "000657ebe", "000659ebe", "000663ebe", "000667ebe", "000673ebe", "000678ebe", "000690ebe", "000695ebe", "000700ebe", 
+    "000701ebe", "000706ebe", "000712ebe", "000725ebe", "000729ebe", "000730ebe", "000734ebe", "000739ebe", "000741ebe", 
+    "000744ebe", "000752ebe", "000754ebe", "000755ebe", "000758ebe", "000781ebe", "000789ebe", "000791ebe", "000794ebe", 
+    "000797ebe", "000801ebe", "000804ebe", "000823ebe", "000827ebe", "000837ebe", "000839ebe", "000847ebe", "000849ebe", 
+    "000858ebe", "000862ebe", "000864ebe", "000868ebe", "000870ebe", "000873ebe", "000874ebe", "000876ebe", "000877ebe", 
+    "000879ebe", "000880ebe", "000891ebe", "000901ebe", "000904ebe", "000906ebe", "000914ebe", "000916ebe", "000923ebe", 
+    "000924ebe", "000928ebe", "000932ebe", "000934ebe", "000938ebe", "000943ebe", "000947ebe", "000949ebe", "000955ebe", 
+    "000956ebe", "000972ebe", "000990ebe", "000991ebe", "000993ebe", "000996ebe", "000997ebe", "000999ebe"]
```
ThomasSchellenbergNextCentury, line 24: These are all expected scenes. Did we need to include their corresponding unexpected scenes?

brianpippin: Just a heads up all the unexpected scenes all already have a score worth of 0, so we don't need to do anything with them.

rartiss55 (approved):

scripts/_0_4_7_2_no_inaccessible_goals.py
rartiss55: The answer is likely no but looking at this again, any need to update the weighted_score (not worth) for the unexpected scenes, since those are still 1 (I think)?

PR #68, opened 2022-01-06
MCS-1020 refactor

deanwetherby:
Draft PR to help with async design discussion


mcs_automated_ingest.py
```diff
@@ -27,58 +31,68 @@
...
+            'file': {'StringValue': str(basename), 'DataType': 'String'},
+            'error': {'StringValue': str(traceback.format_exc()), 'DataType': 'String'}
+        })
+        logging.info(f"Sending {response}")
+
+
 def main():
+
+    mcs_client = MongoClient('mongodb://mongomcs:mongomcspassword@localhost:27017/mcs')
+    dev_client = MongoClient('mongodb://mongomcs:mongomcspassword@localhost:27017/dev')
```
deanwetherby, line 71: Is it ok to make the client once here and pass it to the various sub-modules? Or is it like ray where we need to make a new one each time?

deanwetherby:
How did you feel about using docker/mongo for unit testing? Should we also add moto to mock the aws boto3 calls?

mcs_automated_ingest.py
brianpippin: You can make this client and pass it around, I've done that before with some of the scripts where I passed the client that way.

README.md
```diff
@@ -42,13 +42,13 @@ python -m pip install pymongo
...
 ```
-python mcs_scene_ingest.py --folder ../genScenes/  --eval_name eval-test --type scene
+python mcs_scene_ingest.py --folder ../genScenes/  --eval_name eval-test
 ```
 
 ## Ingest history file (requires scenes already ingested)
 
 ```
-python mcs_scene_ingest.py --folder ../MCS/SCENE_HISTORY/  --eval_name eval-test --type history    --performer "baseline"
+python mcs_scene_ingest.py --folder ../MCS/SCENE_HISTORY/  --eval_name eval-test --performer "baseline"
```
rartiss55, line 51: We're probably getting rid of this anyway, but if we keep it, this should probably be "mcs_history_ingest.py"?

mcs_history_ingest.py
```diff
@@ -0,0 +1,764 @@
...
+            client,
+            db_string)
+        history_item["score"]["scorecard"] = (
+            calc_scorecard(history, scene)
+            if history_item['category'] == 'interactive' else None
+        )
+
+        # Ignore keys during unit tests because we don't want to query mongo.
+        if not ignore_keys:
+            print("Check Scene Keys")
```
rartiss55, line 358: Should this use logging.info instead of print?

deanwetherby: I haven't looked much at mcs_history_ingest yet. Just copied functions directly out of scene_ingest but your point is valid.

brianpippin:
> How did you feel about using docker/mongo for unit testing? Should we also add moto to mock the aws boto3 calls?

I think it's a great idea, the only thing we will have to double check is that it works easily on Jenkins.

mcs_scene_ingest.py
```diff
@@ -29,27 +29,24 @@ def automated_scene_ingest_file(
...
     mongoDB = client[db_string]
 
     scene_item = build_scene_item(file_name, folder, None)
     collection = mongoDB[SCENE_INDEX]
-    check_exists = collection.find(
+    count = collection.count_documents(
         {
             "name": scene_item["name"],
-            "evaluation": scene_item["eval"]
+            "eval": scene_item["eval"]
```
deanwetherby, line 39: Unit tests were failing until I changed this key. We'll need to double-check what key needs to be used here.

tests/test_mcs_scene_ingest.py
```diff
@@ -1,14 +1,111 @@
...
 import mcs_scene_ingest
+import create_collection_keys
 
 from pymongo import MongoClient
 
 TEST_SCENE_FILE_NAME = "test_juliett_0001_01_debug.json"
 # TEST_INTERACTIVE_SCENE_FILE_NAME = "occluders_0001_17_I1_debug.json"
 TEST_FOLDER = "tests"
 
+class TestMcsSceneIngestMongo(unittest.TestCase):
```
deanwetherby, line 16: Split tests into two TestCases. This one uses the docker/mongo approach with a client setup/teardown that isn't needed for the other tests. This can always be rearranged when refactoring or this testclass can easily be removed or skipped if it proves to be problematic.

brianpippin: I like the way this is working to test the actual ingest, and having the other tests for basically testing the structure we are going to ingest, I think that is a good delineation.

mcs_scene_ingest.py
brianpippin: > db.mcs_scenes.distinct("evaluation")
[ null, "Evaluation 3.5 Scenes", "Evaluation 3.75 Scenes" ]
> db.mcs_history.distinct("evaluation")
[ ]

I believe we don't want it to be "evaluation" that must have been a bug of some sort, "eval" is used throughout the UI

create_collection_keys.py
```diff
@@ -20,23 +20,18 @@ def find_collection_keys(index: str, collection_name: str, mongoDB):
...
-    keys_dict = {}
-    keys_dict["name"] = collection_name
-    keys_dict["keys"] = keys
-
+    keys_dict = {'name': collection_name, 'keys': keys}
     collection = mongoDB["collection_keys"]
     result = collection.update_one(
         {"name": collection_name}, {"$set": keys_dict}, True)
-    print(result)
+    # print(result)
```
ThomasSchellenbergNextCentury, line 32: Should these prints be logs instead?

mcs_scene_ingest.py
```diff
@@ -24,107 +19,39 @@
...
-        f for f in os.listdir(
-            folder) if str(f).endswith("." + extension)]
-    history_files.sort()
-    return history_files
+    # DW: TODO functionize (create_evaluation_index)
+    # Add Keys when a new evaluation item is created
+    if create_collection_keys.check_collection_has_key(
+            scene_item["eval"], mongoDB) is None:
+        create_collection_keys.find_collection_keys(
+            SCENE_INDEX, scene_item["eval"], mongoDB)
```
ThomasSchellenbergNextCentury, line 54: @deanwetherby Reminder about this comment to yourself.

ThomasSchellenbergNextCentury:
Once we finalize what other instructions may be needed for preparing development environments to run the unit tests (installing rootless docker, etc.), we should add that info the the README.

ThomasSchellenbergNextCentury:
Part of the original intent for this ticket was expanding on the unit tests, and I'm still concerned about the amount of coverage we current have, especially for mcs_history_ingest.py:
- The existing `build_history_item` unit tests only check for errors, and not whether the returned history_item has the expected properties, given a specific input. There are a lot of if/else statements in this function, so I imagine there are a lot of different cases to test. We'll also likely want to extract the loading-input-from-json-file behavior out of the function in order to test it properly with custom input.
- The existing `process_score` unit tests are only testing intuitive physics scenes, and not interactive or agent scenes.
- No unit tests for complex functions like `add_weighted_cube_scoring` or `build_new_step_object`, though they could probably be covered by tests on other functions like `build_history_item`
- Tests for all of the above using incorrectly-formatted input, to ensure ingest acts the way we want it to (either failing, or returning corrected data, or whatever).

We should consider addressing these issues as part of this PR.

create_collection_keys.py
deanwetherby: Prints were for my debugging purposes. Will be removed on next push.

mcs_scene_ingest.py
deanwetherby: Removing TODO

deanwetherby:
> especially for mcs_history_ingest.py

Right. I split scene out into a new history module but we are discussing another split for scoring which will be a future ticket.

mcs_history_ingest.py
deanwetherby: Is this true now? Is ignore_keys around just to bypass mongo for testing purposes?

ThomasSchellenbergNextCentury: Yes. I put that in because I wanted the unit tests to be async from mongo at the time. Now that we've added mongo as a dependency to more unit tests, and an easy way to download and run mongo automatically using docker, we probably don't need the `ignore_keys` behavior any more.

README.md
rartiss55: Tried this out for history -- looks like --type is still needed/that requirement should be removed.

rartiss55:
Probably not a huge deal since I don't think this is happening on Linux/Jenkins, but I ran the tests locally -- they all passed, but I'm getting a "ResourceWarning: unclosed <ssl.SSLSocket" message. I'm not sure why since it looks like everything's getting cleaned up correctly?

deanwetherby:
> getting a "ResourceWarning: unclosed <ssl.SSLSocket" message

I am getting the same. Looking into it more this morning.

rartiss55 (approved):

brianpippin (approved):
Created a bunch of tickets to cover targeted refactoring for certain areas, think this new testing approach will prove robust and allow for more test coverage in the future.

PR #72, opened 2022-01-18
MCS-841: Database Refactor

brianpippin:
This handles ingesting to specific collections for each eval.  Script to update database to move collections out on their own.

Also handles MCS-1101 to remove old legacy folder ingestion code.

### DIFF
--- a/mcs_scene_ingest.py
+++ b/mcs_scene_ingest.py
@@ -81,7 +81,12 @@ def build_scene_item(file_name: str, folder: str) -> dict:
 
     scene["eval"] = scene["debug"]["evaluation"]
 
-    scene["evalNumber"] = float(re.sub("[^0-9.]", "", scene["eval"]))
+    eval_number_str = re.sub("[^0-9.]", "", scene["eval"])
+    if "." in eval_number_str:
+        scene["evalNumber"] = float(eval_number_str)
+    else:
+        scene["evalNumber"] = int(eval_number_str)
+
     scene["scene_num"] = scene["debug"]["sceneNumber"]
     if "sequenceNumber" in scene["debug"]:
         scene["test_num"] = scene["debug"]["sequenceNumber"]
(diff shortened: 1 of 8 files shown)


scripts/_0_5_0_migrate_collections.py
```diff
@@ -0,0 +1,64 @@
+def migrate_collections(mongoDB):
```
deanwetherby, line 1: Can this function be split into smaller testable functions?

brianpippin: Added some smaller functions to improve readability.

mcs_scene_ingest.py
```diff
@@ -18,10 +18,31 @@
...
+        }
+    )
+
+    if mapping is None:
+        eval_number = float(re.sub("[^0-9.]", "", eval_name))
+        collection_name = "eval_" + (str(eval_number)).replace(".", "_") + "_scenes"
+        collection.insert_one({"name": eval_name, "collection": collection_name})
+        return collection_name
+    else:
+        return mapping["collection"]
```
ThomasSchellenbergNextCentury, line 43: You've got two `get_scene_collection` functions that are very similar. Should the function be extracted to a separate file, and then imported into both mcs_history_ingest and mcs_scene_ingest?

mcs_scene_ingest.py
```diff
@@ -50,20 +72,16 @@ def automated_scene_ingest_file(
...
 
-    if eval_name is None:
-        scene["eval"] = scene["debug"]["evaluation"]
-        eval_name = scene["debug"]["evaluation"]
-    else:
-        scene["eval"] = eval_name
+    scene["eval"] = scene["debug"]["evaluation"]
 
-    scene["evalNumber"] = int(re.sub("[^0-9]", "", scene["eval"]))
+    scene["evalNumber"] = float(re.sub("[^0-9.]", "", scene["eval"]))
```
ThomasSchellenbergNextCentury, line 84: You're making the evalNumber into a float in a few places. I understand that's important to support Eval 3.5, but won't that also make "Eval 4" into "Eval 4.0"? Will that be OK?

PR #73, opened 2022-01-21
MCS-1116:  Script to update saved queries

brianpippin:
MCS-841 needs to be merged before this script, which is just to update saved queries.

### DIFF
--- a/scripts/_0_5_0_1_update_saved_queries.py
+++ b/scripts/_0_5_0_1_update_saved_queries.py
@@ -1,0 +1,31 @@
+field_mapping = {
+    "mcs_history.Evaluation 4 Results": "eval_4_results",
+    "mcs_history.Evaluation 3.75 Results": "eval_3_75_results",
+    "mcs_history.Evaluation 3.5 Results": "eval_3_5_results",
+    "mcs_history.Evaluation 3 Results": "eval_3_results",
+    "mcs_history.Evaluation 2 Results": "eval_2_results",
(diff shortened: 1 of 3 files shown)


scripts/_0_5_0_1_update_saved_queries.py
```diff
@@ -0,0 +1,31 @@
...
+def update_saved_queries(mongoDB):
+    saved_queries_collection = mongoDB["savedQueries"]
+    all_documents = saved_queries_collection.find({})
+    
+    for document in all_documents:
+        # Update the saved tabs object fieldTypes
+        for tab in document["user"]["queryBuilderState"]["queryTabs"]:
+            for tab_query_obj in tab["tabQueryObj"]:
+                if tab_query_obj["fieldType"] in field_mapping:
+                    tab_query_obj["fieldType"] = field_mapping[tab_query_obj["fieldType"]]
```
deanwetherby, line 23: This amount of nesting usually gives me pause but since this is a run-once migration script I won't worry about it.

deanwetherby (approved):

scripts/_0_5_0_1_update_saved_queries.py
brianpippin: Yes I don't like that must nesting but it has to do with the object state that is stored in the database and just trying to update those saved query states.  Seemed like the quickest, fastest way to update this.  This script runs in under a minute with the prod db and I checked that it's working so I agree that it's not ideal but okay for a run once script.

PR #76, opened 2022-03-14
UI Ignores Spatial Elimination Scenes A3, C3, A4, C4 in Results

jaudick:


mcs_history_ingest.py
```diff
@@ -483,6 +484,16 @@ def add_weighted_cube_scoring(history_item: dict, scene: dict) -> tuple:
...
+        elif scene["goal"]["sceneInfo"]["tertiaryType"] == "spatial elimination":
+            if history_item["scene_goal_id"] in \
+                    SPATIAL_ELIMINATION_IGNORED:
+                weighted_score = 0
+                weighted_score_worth = 0
+                weighted_confidence = 0
+            else:
+                weighted_score = history_item["score"]["score"]
+                weighted_score_worth = 1
+                weighted_confidence = history_item["score"]["confidence"]
```
brianpippin, line 496: @jaudick - I know you did what we discussed but I also think we can take this opportunity to clear up this function, I was thinking we could do something that would improve readability, so anywhere that is an if or elif, we change it to return (0, 0, 0) or something like return (history_item["score"]["score" * 2, 2, float(score*2), and then delete all the else blocks, and change the last else statement to return (history_item["score"]["score]", 1, history_item["score"]["confidence"]) this would help us to eliminate code duplication, and potentially clean up the code, thoughts?

jaudick: I'm definitely support that change, I'll make some tweaks

brianpippin: Jacob that function looks so much cleaner now, refactoring for the win!

brianpippin (approved):

ThomasSchellenbergNextCentury (approved):
I haven't tried running the code myself, but it LGTM otherwise.

PR #81, opened 2022-04-20
Mcs 978 object id repeated failed

cedorman:
Changed to use obj ids rather than locations.  Note:  This is based off of MCS-1252, so that should be reviewed first.


ThomasSchellenbergNextCentury:
I changed the base of this PR to branch MCS-1252, to simplify the diffs. Hope you don't mind.

PR #82, opened 2022-04-22
Mcs 978 object id repeated failed

cedorman:
Use the object ids rather than locations where possible in the scorecard, since MCS-977 is reporting the object ids for containers and objects.  This allows us to count which objects are getting either failures or repeated openings, which changes the resulting output scorecard format.


ThomasSchellenbergNextCentury (approved):
Looks good, but do we need to update any of the info in the scorecard README?

PR #83, opened 2022-05-03
Mcs 1243 ramp traversal

cedorman:
Implemented MCS 1243, which keeps track of ramp traversals.  This counts:

- number of times went up
- number of times went down
- number of times started going up and abandoned
- number of times started going down and abandoned
- number of times fell off


ThomasSchellenbergNextCentury:
@cedorman Were you going to add a description for the new metric in the README?

ThomasSchellenbergNextCentury (changes requested):
I'm getting a unit test error:

```
======================================================================
ERROR: test_on_ramp (tests.test_scorecard.TestMcsScorecard)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/tschellenberg/Development/darpa/mcs-ingest/tests/test_scorecard.py", line 379, in test_on_ramp
    on_ramp_bool, ramp_id = scorecard.on_ramp(position)
ValueError: too many values to unpack (expected 2)
```

It looks like `on_ramp` returns 3 outputs, but it's only expecting 2 here?

scorecard/scorecard.py
```diff
@@ -576,6 +594,182 @@ def calc_not_moving_toward_object(self):
...
+                        'went_down_abandoned': 0,
+                        'ramp_fell_off': 0}
+        last_ramp_action_step = 0
+        last_ramp_action_position = old_position
+        last_ramp_action = 'None'
+
+        for single_step in steps_list:
+            step = single_step['step']
+            if step == 246:
+                print("")
```
ThomasSchellenbergNextCentury, line 620: ?

scorecard/scorecard.py
```diff
@@ -576,6 +594,182 @@ def calc_not_moving_toward_object(self):
...
+            continue
+
+        self.ramp_actions = {}
+        self.ramp_actions.update(ramp_actions)
+        logging.debug('Ending calculating ramp actions')
+        return self.ramp_actions
+
+    def on_ramp(self, position) -> (bool, str):
+        '''Determine if a position is in a ramp.  Return
+        a boolean and, if True, the ID'''
```
ThomasSchellenbergNextCentury, line 737: I think this function's return type and docstring need to be updated.

scorecard/scorecard_location_utils.py
```diff
@@ -0,0 +1,170 @@
...
+
+    corners = get_corners_from_center_size_rotation(center, size, rotation)
+    is_actually_on = is_point_in_polygon(pt, corners)
+    return is_actually_on
+
+    # It is not clear if the following is needed.  If it is, then
+    # add size_limit to the parameters and uncomment below.
+    #     is_near_base = is_point_near_base(pt, center, size, rotation,
+    #            size_limit)
+    #     return is_actually_on or is_near_base
```
ThomasSchellenbergNextCentury, line 170: Can you elaborate here?

scorecard/scorecard.py
```diff
@@ -576,6 +594,182 @@ def calc_not_moving_toward_object(self):
...
+                      old_position,
+                      new_position) -> bool:
+        # If they dropped a lot, then must have fallen
+        amount_down = new_position['y'] - old_position['y']
+        if amount_down < FALL_DISTANCE:
+            return True
+
+        # Might need to calculate whether went over the a side??
+        # Complicated because they might be 'cutting the corner'
+        # when going down ramp....
```
ThomasSchellenbergNextCentury, line 769: What are the pros and cons of this?

scorecard/scorecard.py
```diff
@@ -260,11 +274,15 @@ def __init__(self, history: dict, scene: dict):
...
-        self.calc_attempt_impossible()
+
         self.calc_open_unopenable()
         self.calc_relook()
         self.calc_revisiting()
         self.calc_not_moving_toward_object()
+        self.calc_ramp_actions()
+
+        # To be implemented
+        # self.calc_attempt_impossible()
```
brianpippin, line 285: Is this for another ticket where this will be implemented?

scorecard/scorecard.py
cedorman: Weird, don't know how it worked on my machine.  Fixed.

scorecard/scorecard.py
cedorman: Dang.  Hate when I leave debugging stuff...

scorecard/scorecard_location_utils.py
cedorman: I improved the comment here, for below comment, and the README.

scorecard/scorecard.py
cedorman: Pro:   possibly more accurate counting in edge cases when they do something that we don't catch.  
Con:  really more complicated geometry and logic steps.  

It's a balance between making the logic even more complicated than it already is and having the mathematically 'right'
answer.  I'm convinced that the current logic (it fell!) works in 99% of the cases.  However, I have a sneaking suspicion 
that something could happen where a fall would count or we would miss one.   But I think that doing the geometry / logic
in those cases is not worth time / effort, and I think that I might be adding more logic holes than filling in.

scorecard/scorecard.py
cedorman: MCS-860.  Note that this is a catch-all ('it didn't work') and we should be much more explicit about what we want to include.  It would be easy to count all the things that don't count as 'success' if that is what is wanted.

cedorman:
Updated all the comments, fixed the unit test bug.

ThomasSchellenbergNextCentury (approved):

scorecard/scorecard_location_utils.py
```diff
@@ -0,0 +1,177 @@
...
+    is_inside = False
+    for p1, p2 in zip(polygon, polygon[1:]):
+        if ((p1.y > pt.y) != (p2.y > pt.y) and
+                pt.x < (p2.x - p1.x) * (pt.y - p1.y) / (p2.y - p1.y) + p1.x):
+            is_inside = not is_inside
+
+    return is_inside
+
+
+def rotate_x_z(x, z, center_x, center_z, rotation):
```
brianpippin, line 65: This function doesn't type the variables like those around it.

cedorman: Fixed a couple of functions, to be consistent.

PR #85, opened 2022-05-18
Mcs 1241 which side of platform

cedorman:
Return whether the agent went on the correct side of the platform or not, based on matching the side that targetSide is set to.


ThomasSchellenbergNextCentury (commented):
Please update the README with a description for this new scorecard metric. Otherwise, everything looks good to me.

scorecard/scorecard.py
```diff
@@ -820,6 +823,37 @@ def calc_tool_usage(self):
...
+                x = output['position']['x']
+                if x < 0 and target_side == 'left':
+                    self.correct_platform_side['correct_side'] = True
+                else:
+                    self.correct_platform_side['correct_side'] = False
+                if x > 0 and target_side == 'right':
+                    self.correct_platform_side['correct_side'] = True
+                else:
+                    self.correct_platform_side['correct_side'] = False
+            old_y = new_y
```
rartiss55, line 854: I realized that for support relations (which is still in review https://github.com/NextCenturyCorporation/mcs-private/pull/239) "center" is the correct side for the target. Does that need to be taken into account here? I believe the y wouldn't change in that case.

rartiss55, line 854: I realized that for support relations (which is still in review https://github.com/NextCenturyCorporation/mcs-private/pull/239) "center" is the correct side for the target. Does that need to be taken into account here? I believe the y wouldn't change in that case.

ThomasSchellenbergNextCentury: There's a separate ticket for checking the correct door in the scorecard (MCS-1244) which will be labeled with a different tag (`correctDoor`).

scorecard/README.md
```diff
@@ -153,6 +154,14 @@ object to achieve a goal.  This counts the different types of manipulation,
 inlcuding pushing, pulling, rotating, torquing, and moving. It also counts 
 the number of times that they attempted to do so and failed.
 
+#### Platform Side Correctness
+
+For several task types, Agent Identification, Spatial Elimination, Interactive 
+Object Permanence, the agent needs to move off of the platform to one side or 
+the other.  This element of the scorecard determines whether the agent moved to
+the correct side.
```
ThomasSchellenbergNextCentury, line 162: I'm recommend being a bit more descriptive, like: "For several task types that use a forced-choice paradigm with a platform bisecting the room, like Agent Identification or Spatial Elimination, ..."

ThomasSchellenbergNextCentury (approved):

rartiss55 (approved):

PR #89, opened 2022-05-25
MCS-1244 - correct_door for scorecard

rartiss55:
Used for "doorcluder" tasks to determine if correct door was chosen


scorecard/scorecard.py
```diff
@@ -287,14 +288,17 @@ def score_all(self) -> dict:
...
+        self.calc_correct_door_opened()
 
         # To be implemented
         # self.calc_attempt_impossible()
 
         return {
             'repeat_failed': self.repeat_failed,
             'attempt_impossible': self.attempt_impossible,
+            'correct_door_opened': self.correct_door_opened,
+            'correct_platform_side': self.correct_platform_side,
```
rartiss55, line 301: Made minor updates as well for the platform_side stuff done prior to this ticket, since it wasn't actually getting returned in the scorecard object.

scorecard/scorecard.py
```diff
@@ -854,6 +858,46 @@ def calc_correct_platform_side(self):
             old_y = new_y
         return self.correct_platform_side
 
+    def calc_correct_door_opened(self):
+        '''Determine if the ai agent went through the correct door'''
```
rartiss55, line 862: So I just copied the data structure used for platform_side, but do we want this simplified? Currently its something like this: 

{'correct_platform_side': {'correct_side': False})

Couldn't we just do a boolean and make it T/F/None for both?

scorecard/scorecard.py
```diff
@@ -854,6 +858,46 @@ def calc_correct_platform_side(self):
...
+
+        # Does this scene have a correctDoor? If not, return empty dict.
+        goal = self.scene.get('goal')
+        if 'sceneInfo' in goal and 'correctDoor' in goal['sceneInfo']:
+            correct_door = goal['sceneInfo']['correctDoor']
+        else:
+            return
+
+        self.correct_door_opened = defaultdict(bool)
```
rartiss55, line 871: So I just copied the structure used for platform_side, but do we want this simplified? Currently its something like this:

{'correct_platform_side': {'correct_side': False})

Couldn't we just do a boolean and make it T/F/None for both?

kdrumm:
I'm having unit test failures although I'm not sure what could have caused those.

rartiss55:
> I'm having unit test failures although I'm not sure what could have caused those.

Ah, I see them too. I made a small change and didn't rerun them -- I'll take a closer look.

scorecard/scorecard.py
```diff
@@ -854,6 +858,46 @@ def calc_correct_platform_side(self):
             old_y = new_y
         return self.correct_platform_side
 
+    def calc_correct_door_opened(self):
+        '''Determine if the ai agent went through the correct door'''
+
+        # Does this scene have a correctDoor? If not, return empty dict.
```
brianpippin, line 864: Should this return null instead of an empty dict then if we are getting rid of it returning a dictionary.

scorecard/scorecard.py
ThomasSchellenbergNextCentury: Just checking - It looks like you also updated the unit test for the platform_side - Correct? (If not, we should do that.)

scorecard/scorecard.py
ThomasSchellenbergNextCentury: I can't think of any reason why we'd want to expand this metric in the future, so I agree in simplifying the dict structure. We should probably do the same for platform_side, and check the other new metrics as well. Pinging @cedorman in case he has any concerns.

scorecard/scorecard.py
```diff
@@ -854,6 +858,46 @@ def calc_correct_platform_side(self):
...
+                    if 'shows' in obj and len(obj['shows']) > 0:
+                        door_x_pos = obj['shows'][0]['position']['x']
+                        if door_x_pos == 0 and correct_door == 'center':
+                            self.correct_door_opened['correct_door'] = True
+                        elif door_x_pos < 0 and correct_door == 'left':
+                            self.correct_door_opened['correct_door'] = True
+                        elif door_x_pos > 0 and correct_door == 'right':
+                            self.correct_door_opened['correct_door'] = True
+                        else:
+                            self.correct_door_opened['correct_door'] = False
```
ThomasSchellenbergNextCentury, line 897: Not to call you out specifically, since this happens throughout the scorecard, but I think it would be helpful to compile a list of all the assumptions we make for each metric, maybe in the docstring for each `calc_*` function. For example: door object IDs begin with door_; door positions are either position X, negative X, or zero; scene has "correctDoor" tag set to either "left", "right", or "center".

How does that sound?

scorecard/scorecard.py
rartiss55: This should return itself/None now if not initialized.

scorecard/scorecard.py
rartiss55: It looks like the other calc_* functions have better comments throughout, so I added assumptions just for doors/platform sides.

scorecard/scorecard.py
rartiss55: Yes 👍

PR #90, opened 2022-06-02
MCS-1355 - Fix ValueError for weighted_confidence

rartiss55:
In the history file, we accept the None type as well as a float for confidence/score values. When the weighted_confidence is calculated, ingest attempts cast to a float(), which was failing because of trying to convert "None". In that case, we want to just return None/null for weighted_confidence. Re-ingested and checked the UI locally with this change, didn't look like anything broke + added some unit tests as well. 

Note that this was to fix weighted_confidence, the original confidence value is left as is.

### DIFF
--- a/tests/test_mcs_history_ingest.py
+++ b/tests/test_mcs_history_ingest.py
@@ -347,16 +347,14 @@ class TestMcsHistoryIngest(unittest.TestCase):
             "goal": {
                 "sceneInfo": {
                     "tertiaryType": "spatial elimination"
-                },
-                'answer': {
-                    'choice': 'plausible'
                 }
             }
         }
 
         history_item = {
-            'category': 'passive',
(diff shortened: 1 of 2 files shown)


tests/test_mcs_history_ingest.py
```diff
@@ -347,16 +347,14 @@ def test_ignored_spatial_elimination_scenes(self):
...
-                'answer': {
-                    'choice': 'plausible'
                 }
             }
         }
 
         history_item = {
-            'category': 'passive',
-            'test_type': 'spatial elimination',
+            'category': 'interactive',
```
rartiss55, line 355: Cleaned up a previous test a bit to more accurately reflect history file contents.

brianpippin (approved):
LGTM

PR #92, opened 2022-06-13
Count the number of times the performer tried to pickup an object that is not pickupable

jaudick:
I ran human input mode with my custom scene and ran around pickup stuff up and dropping object. I also interacted with an agent and picked up its held object too. I tried picking up the same object 1 time, 2 times, or 3 times to add variation.


scorecard/scorecard.py
```diff
@@ -968,6 +974,40 @@ def get_distance_from_path(self, start_pos: Dict[str, float],position: Dict[str,
...
+        failed_objects = defaultdict(int)
+
+        for single_step in steps_list:
+            step = single_step['step']
+            action = single_step['action']
+            output = single_step['output']
+            if action == 'PickupObject':
+                return_status = output['return_status']
+                if return_status in ["SUCCESSFUL"]:
+                    obj_id = get_relevant_object(output)
```
brianpippin, line 998: You call obj_id = get_relevant_object in the if and the elif, would it make sense to just calculate it above the if statement once?

brianpippin:
The logic looks good, just worried about all the additional code that seems to be there for debugging compared to the minimal that might be needed just in the function.  Seems like all you need is the NOT_PICKUPABLE check and the +1 and it should be a very simple function.  All the obj_id stuff adds unneeded complexity I think.

scorecard/scorecard.py
```diff
@@ -968,6 +974,40 @@ def get_distance_from_path(self, start_pos: Dict[str, float],position: Dict[str,
...
-            
+    
+    def calc_pickup_not_pickupable(self):
+        ''' 
+        Determine the number of times that the performer tried to
+        pickup an object than cannot be picked up:
+        Agents, Blobs, Floors, Walls, Platforms, Platform Lips, Ramps,
+        Static objects (sofas, chairs, etc..), Tools, Walls
+        '''
+        logging.debug('Starting calculating unpickupable')
```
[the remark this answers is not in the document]
jaudick: I modeled this after `calc_open_unopenable()` that had a almost identical logic but also added additional code for logging specific id's. All of the other funcitons with `calc` appear to contain logging as well. Do you want to maintain this logging design for new scorecard checks?

scorecard/scorecard.py
jaudick: Yes that will work too

jaudick:
@brianpippin I pushed a new commit with the logging removed and more concise logic. What do you think?

ThomasSchellenbergNextCentury (approved):

brianpippin (approved):

PR #93, opened 2022-06-13
Count number of times performer interacts with non-agents

jaudick:


brianpippin (approved):

rartiss55:
The scorecard.py code looks fine to me. I think there should be a bit describing the non-agent interactions in the README like we have for the other scorecard values. Speaking of the README, I have a question about testing:

I’ve seen varying ways of addressing tests on recent PRs so perhaps this is a larger discussion, but for testing, did we still want to follow what’s in the scorecard README for testing: a way to generate a history file again and recommit in case those outputs change + track the ground truth, or is what’s here sufficient/does that only apply to certain scenarios? Readme link here: https://github.com/NextCenturyCorporation/mcs-ingest/blob/master/scorecard/README.md#testing-the-scorecard and an example of tests here: https://github.com/NextCenturyCorporation/mcs-ingest/pull/85.

brianpippin:
I think the README needs to be updated because he totally nixed needing to generate file that way and all the files are included for testing.  I believe that whole section is deprecated at this point.

PR #97, opened 2022-06-29
Count times walked into walls, ramp sides, occluder, platform walls but not lips

jaudick:


scorecard/scorecard.py
```diff
@@ -346,6 +351,9 @@ def get_interact_with_non_agent(self):
     def get_pickup_not_pickupable(self):
         return self.pickup_not_pickupable
 
+    def get_walked_into_structures(self):
```
rartiss55, line 354: Something I just noticed: doesn't hurt to have them, but do we actually use any of these scorecard get_* methods anywhere?

scorecard/scorecard.py
```diff
@@ -1019,4 +1026,122 @@ def calc_interact_with_non_agent(self):
...
+        bounding_boxes = [
+            {'id': struct['id'], 'bounding_box': struct['shows'][0]['boundingBox']}
+            for struct in structures]
+        ramp_bounding_boxes = [
+            {'id': struct['id'], 'bounding_box': struct['shows'][0]['boundingBox']}
+            for struct in structures if struct['id'].startswith('ramp')]
+        room_dimensions = self.scene['roomDimensions']
+        
+        # Keeps track of obstruction ids, this is not being used now but may be useful
+        obstructions = []
```
rartiss55, line 1108: Did we want to save these ids somewhere in the actual scorecard object?

scorecard/scorecard.py
```diff
@@ -1019,4 +1026,122 @@ def calc_interact_with_non_agent(self):
...
+        walked_into_structures = 0
+        structures = [obj for obj in self.scene['objects']
+            if obj.get('structure') is True]
+        bounding_boxes = [
+            {'id': struct['id'], 'bounding_box': struct['shows'][0]['boundingBox']}
+            for struct in structures]
+        ramp_bounding_boxes = [
+            {'id': struct['id'], 'bounding_box': struct['shows'][0]['boundingBox']}
+            for struct in structures if struct['id'].startswith('ramp')]
+        room_dimensions = self.scene['roomDimensions']
```
ThomasSchellenbergNextCentury, line 1105: Some older scenes don't have a `roomDimensions`. In this case, you can fall back to using the default value:

`{'x': 10, 'y': 3, 'z': 10}`

scorecard/scorecard.py
```diff
@@ -1019,4 +1026,122 @@ def calc_interact_with_non_agent(self):
...
\ No newline at end of file
+        return self.interact_with_non_agent
+
+    def get_min_max_bounding_box_coords(self, bounding_box, key):
+        minimum = bounding_box[0][key]
+        maximum = bounding_box[0][key]
+        for i in range(1, len(bounding_box)):
+            value = bounding_box[i][key]
+            minimum = value if value < minimum else minimum
+            maximum = value if value > maximum else maximum
```
ThomasSchellenbergNextCentury, line 1037: FYI Python has built-in `min` and `max` functions, so in the future you could just do `minimum = min(value, minimum)`

brianpippin (approved):
We definitely need to do MCS-1114 after the evaluation, the score card is getting too large.  We need a refactor to figure out what functions can be common, and then great a module for each parameter to decrease the complexity of this file.  This code however looks good.

scorecard/scorecard.py
rartiss55: Sounds like just having the count is fine for now then.

rartiss55 (approved):

PR #99, opened 2022-07-22
Fixed one issue with the correct platform side scorecard metric.

ThomasSchellenbergNextCentury:
~~There may be other fixes needed beyond the `if` statement.~~

EDIT: I think I fixed all of the issues.


brianpippin:
We need to discuss if we need to write a script to do this, or re-ingest all the history files

ThomasSchellenbergNextCentury:
@brianpippin This is ready to go now. I think you can ignore the changes to the default value (None vs. False) for your mongo script; the important part is in the `if` statement.

brianpippin (approved):

brianpippin:
Merging this, the script will be a different entity.

PR #100, opened 2022-07-22
Script to rescore platform side

brianpippin:


scripts/_0_5_7_1_rescore_platform_side.py
```diff
@@ -0,0 +1,17 @@
...
+
+def rescore_platform_side(mongoDB):
+    results_collection = mongoDB["eval_5_results"]
+    scenes_collection = mongoDB["eval_5_scenes"]
+
+    history_files = results_collection.find({"category": "interactive"})
+
+    for history in history_files:
+        scene = scenes_collection.find_one({"name": history["name"]})
+        scorecard = Scorecard(history, scene)
```
rartiss55, line 11: Tried this locally and couldn't get it working -- realized something: are we actually saving the position per step of the performer in the history record in mongo? I think it works in the scorecard because that's just using the history file as is prior to ingest. 

If the answer to that is no, I think we can just do some form of download the history files that need an update + re-ingest instead of running again, right?

rartiss55: Or we can even directly reference the history files themselves  in this script maybe?

rartiss55: Trying to think of what this affects:
- agent identification
- IOP
- solidity
- spatial elimination
- support relations

brianpippin: @rartiss55 - I have fixed this to use this history file from S3, let me know what you think now.  I ran it on some results and it seemed to work for me.

rartiss55 (approved):
@brianpippin Tested with a handful of scenes locally, LGTM!

PR #104, opened 2022-08-25
MCS-1443: Add ability to track successful agent interactions in the scorecard

brianpippin:


rartiss55 (approved):

scorecard/scorecard.py
```diff
@@ -1028,8 +1035,12 @@ def calc_interact_with_non_agent(self):
                 resolved_obj = output['resolved_object']
                 if resolved_obj != '' and resolved_obj not in agents:
                     interact_with_non_agent += 1
+                if resolved_obj != '' and resolved_obj in agents and (
+                        output['return_status'] == "SUCCESSFUL"):
+                    interact_with_agent += 1
 
         self.interact_with_non_agent = interact_with_non_agent
+        self.interact_with_agent = interact_with_agent
         return self.interact_with_non_agent
```
rartiss55, line 1044: I suppose it doesn't matter too much since the return types aren't used anymore, but should the return type be removed here? Or maybe add both properties?

brianpippin: I am actually thinking we should go and remove all the return types in the ticket to refactor the score card, nothing uses any of the returns in all these functions.

PR #106, opened 2022-09-21
MCS-1422:  Add indexes automatically when running a new evaluation 

brianpippin:
Once there is one item in the new collection.  A collection does not exist until it has at least one item in it.

### DIFF
--- a/mcs_history_ingest.py
+++ b/mcs_history_ingest.py
@@ -30,6 +30,7 @@ TEAM_MAPPING_DICT = {
 
 # Convert Eval names used to 'pretty' scene record names for UI
 EVAL_SCENE_MAPPING_DICT = {
+    "eval_3": "Evaluation 3 Scenes",
     "eval_3-5": "Evaluation 3.5 Scenes",
     "eval_3-75": "Evaluation 3.75 Scenes",
     "eval_4": "Evaluation 4 Scenes",
@@ -40,6 +41,7 @@ EVAL_SCENE_MAPPING_DICT = {
 }
 # Convert Eval names used to 'pretty' history record names for UI
 EVAL_HIST_MAPPING_DICT = {
+    "eval_3": "Evaluation 3 Results",
     "eval_3-5": "Evaluation 3.5 Results",
     "eval_3-75": "Evaluation 3.75 Results",
     "eval_4": "Evaluation 4 Results",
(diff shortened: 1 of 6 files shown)


mcs_scene_ingest.py
```diff
@@ -22,6 +22,31 @@
 SCENE_DEBUG_EXTENSION = "_debug.json"
 
 
+def copy_indexes(db_string: str,
+        client: MongoClient,
+        new_collection_name: str) -> None:
+    mongoDB = client[db_string]
+    print("copy indexes")
```
ThomasSchellenbergNextCentury, line 29: Is this print needed?

tests/test_mcs_scene_ingest.py
```diff
@@ -10,7 +10,7 @@
 from pymongo import MongoClient
 
 TEST_SCENE_FILE_NAME = "test_data/test_juliett_0001_01_debug.json"
-# TEST_INTERACTIVE_SCENE_FILE_NAME = "occluders_0001_17_I1_debug.json"
```
ThomasSchellenbergNextCentury, line 13: Is this commented-out line needed?

mcs_history_ingest.py
```diff
@@ -93,6 +95,31 @@
...
+    else: 
+        return
+
+    # Get the collection for the current eval using new name
+    # Copy the indexes one by one to new collection from last eval
+    new_results_collection = mongoDB[new_collection_name]
+    scene_indexes = old_collection.index_information()
+    for key in scene_indexes:
+        index_tuple = scene_indexes[key]["key"][0]
+        new_results_collection.create_index([(index_tuple[0], int(index_tuple[1]))])
```
ThomasSchellenbergNextCentury, line 120: Is this the same code that's also used in `mcs_scene_ingest`? Would it make sense to move this function to a separate file, so `mcs_history_ingest` and `mcs_scene_ingest` can both import the same function?

mcs_history_ingest.py
```diff
@@ -564,6 +591,10 @@ def automated_history_ingest_file(
     collection_name = get_history_collection(db_string, client, history_item["eval"])
 
     collection = mongoDB[collection_name]
+    total_documents = collection.count_documents({})
+    if total_documents == 1:
+        copy_indexes(db_string, client, collection_name)
```
ThomasSchellenbergNextCentury, line 596: Why might `total_documents` equal 1?

brianpippin: When 1 document has been ingested and the collection created, so after we create an evaluation, we copy the indexes.

tests/test_mcs_scene_ingest.py
brianpippin: Yes as it was just uncommented out for the tests.

mcs_history_ingest.py
brianpippin: At this point there are many functions that history and scene both use and have been duplicated, we have decided to punt on refactoring ingest, also some of the functions are just slightly different as well.

brianpippin: We could potentially add a refactor ticket to create a shared file between them and try to do some refactoring though if there is the appetite to bite that work off, it might be worth it.

brianpippin: @ThomasSchellenbergNextCentury I decided it was probably time to refactor the now three duplicate functions that were in both scene and history ingest.  Not so much a refactor as a cleanup so just moved the three functions that were duplicate into a common file.

ThomasSchellenbergNextCentury: Glad it turned out to be fairly easy!

ThomasSchellenbergNextCentury (approved):
Tests pass

PR #109, opened 2023-02-22
MCS-1673: Script to add domain type to previous evaluation items

brianpippin:


rartiss55 (approved):
Was able to run the script/LGTM!

scripts/_0_6_5_add_domain_type.py
```diff
@@ -0,0 +1,86 @@
...
+
+# Some older evals had _ in between the words (hence while it appears
+#   to have duplicates)
+passive_objects = [
+    "object_permanence",
+	"shape_constancy",
+	"spatio_temporal_continuity",
+    "object permanence",
+	"shape constancy",
+	"spatio temporal continuity",
```
ThomasSchellenbergNextCentury, line 15: Not sure if it matters, but these are listed twice.

ThomasSchellenbergNextCentury (approved):

scripts/_0_6_5_add_domain_type.py
brianpippin: I added a comment above it Thomas, the database has both "shape_constancy" and "shape constancy" in it, so that's why I need to check for both.  I think the underscore one is from eval 2 and 3 perhaps.  Either way the UI cleans up both strings, but if you look on line 7 there is a comment about it.

ThomasSchellenbergNextCentury: Whoops! I totally missed that. Thanks!

PR #112, opened 2023-03-02
MCS-1597 Added pickup_non_target scorecard metric.

ThomasSchellenbergNextCentury:
UI PR coming soon!


jaudick (approved):
It works and looks good. Question though, i know for all of my scenes I have been using debug scenes so its easier for everyone else to ingest it in the UI. Is there a reason you are using a non debug scene?

scorecard/scorecard.py
```diff
@@ -1003,6 +1009,43 @@ def get_distance_from_path(self, start_pos: Dict[str, float],position: Dict[str,
             p1 = p2
         return dist
     
+    def calc_pickup_non_target(self):
+        """
+        Calculate whether the performer agent picked up a non-target
```
rartiss55, line 1014: Should there be something like "if not tool choice, return" here? Thinking about the ambiguous multi retrieval scenes.

scorecard/scorecard.py
```diff
@@ -1003,6 +1009,43 @@ def get_distance_from_path(self, start_pos: Dict[str, float],position: Dict[str,
...
+        ]
+        if target_list and soccer_ball_list:
+            for step_data in self.history['steps']:
+                # Identify a successful pickup
+                if (
+                    step_data['action'] == 'PickupObject' and
+                    step_data['output']['return_status'] == "SUCCESSFUL"
+                ):
+                    # Use "get" for backwards compatibility with old histories
+                    resolved_id = step_data['output'].get('resolved_object')
```
rartiss55, line 1039: could use "get_relevant_object(output)" here/in the future.

rartiss55:
> It works and looks good. Question though, i know for all of my scenes I have been using debug scenes so its easier for everyone else to ingest it in the UI. Is there a reason you are using a non debug scene?

This is a good question -- I've done both in the past but I think we should stick to regular, non-debug scene files -- there's a lot more new scenes/tests so I'm a bit more wary of pushing debug scenes to this repo now.

ThomasSchellenbergNextCentury:
@jaudick @rartiss55 Oh, doh, I JUST pushed a commit to replace my non-debug test scene with a debug version...

As far are debug vs. non-debug goes, I don't have a strong opinion. Both concerns are valid. Does anyone else feel strongly about either option?

scorecard/scorecard.py
ThomasSchellenbergNextCentury: I'd prefer to keep the metric as generic as possible. That being said, you're right, it doesn't make sense to use for ambiguous multi-retrieval scenes. I've added a check to ignore such scenes.

jaudick:
@ThomasSchellenbergNextCentury @rartiss55 I like the debug scenes because I also use these scenes for the ui ingesting and I don't think there are negatives to having the debug scenes. I don't remember but for evals do we ingest the debug scenes or non-debug scenes?

rartiss55:
> @ThomasSchellenbergNextCentury @rartiss55 I like the debug scenes because I also use these scenes for the ui ingesting and I don't think there are negatives to having the debug scenes. I don't remember but for evals do we ingest the debug scenes or non-debug scenes?

@jaudick We ingest debug scenes/ingest won't work properly without debug scenes. My concern about the debug files was that there is info in there we may not want exposed before the evaluation itself (mostly a concern for new scene types) and this repo is public. I will go with whatever the team decides though.

rartiss55 (approved):
Code LGTM!

meant to note here if using the debug file/ we plan to keep those + use them to test via ingest, the one attached to this PR needs a domainType in sceneInfo (I think its "interactive objects"). Also I matched the scene name to match the one in the history files, and updated the eval properties in there for ingestion (evaluation_name and team).

ThomasSchellenbergNextCentury:
Hmm... there are pros and cons for each. I'm thinking the convenience of having the debug files for testing outweighs the likelihood that the TA1s will be digging through this repo and actually gaining useful information from looking at the debug files? So I'm leaning toward keeping the debug files in here for now.

brianpippin:
Reviewing this code now but wanted to chime into the debug vs non debug.  I actually think the pros of using debug files (which is what we really ingest and ingest won't work without), is better in the long run (just in case we miss something that a debug file causes).  At this stage I am far less concerned with teams "cheating" on the test, especially since the tests are fun with non debug files so even knowing some fields exist doesn't mean they will be in the test files.  Anyways, including debug files has my full support.

PR #113, opened 2023-03-02
MCS-1596 which door is opened

jaudick:
Note that the specifics of shell game set rotation are in progress after our discussion with the psychs


brianpippin (approved):
Code and Tests look good.

ThomasSchellenbergNextCentury (approved):

PR #116, opened 2023-03-07
MCS-1593 - multi reward: number of rewards achieved, address ambiguous scenes

rartiss55:
Number of rewards achieved will be tracked for interactive scenes (not just multi retrieval) as discussed. So it should be 0-n for interactive, and None/null for all other scenes. Note that this isn't needed in the roll up/test overview page, so there's no UI PR. 

Also added fixes to address ambiguous multi retrieval scenes.


mcs_history_ingest.py
```diff
@@ -487,6 +509,20 @@ def build_new_step_obj(
...
             interactive_goal_achieved = 1
 
+        # For now, simply check if objects were picked up or dropped, 
+        # since soccer_balls are the only thing pickupable in these
+        # scenes
+        if(multi_retrieval_scoring_override):
+            if(new_step["action"] == "PickupObject" and 
+               output["return_status"] == "SUCCESSFUL"):
+                multi_retrieval_rewards_picked_up += 1
+            if(new_step["action"] == "DropObject" and
```
rartiss55, line 519: Not sure we even care about DropObject, but for these scenes, I realized the only things pickupable are soccer balls, so I'm just checking for successful Pickups/Drops and not worrying about object ids at the moment (but I can check for that if anyone is concerned). I'm also not over writing the value of "reward" that comes from the python API, but just making sure the scene is scored correctly if that makes sense.  

Is there a follow on ticket to move this or investigate moving this to the Python API later? I can reference it here or make a ticket.

ThomasSchellenbergNextCentury: The tickets are MCS-1707 (python api) and MCS-1708 (scene gen).

I think this will work fine. If there are 2 target objects, and they pick up 2 objects, then they must be successful, right? Even in subtraction scenes (like, 4-2 vs 2), they aren't able to reach the subtracted balls, so this will work.

tests/test_scorecard.py
```diff
@@ -566,6 +574,77 @@ def test_number_of_times_walked_into_platform_lips(self):
...
+
+    def test_number_of_rewards_achieved_ambiguous_right(self):
+        scene_file = mcs_scene_ingest.load_json_file(
+            TEST_FOLDER, TEST_SCENE_NUM_REWARDS_AMB)
+        history_file = mcs_scene_ingest.load_json_file(
+            TEST_FOLDER, TEST_HISTORY_NUM_REWARDS_AMB_R)
+        scorecard = Scorecard(history_file, scene_file)
+        scorecard.calc_num_rewards_achieved()
+        # should be 1, since scene is ambiguous
+        assert scorecard.get_number_of_rewards_achieved() == 0
```
ThomasSchellenbergNextCentury, line 646: Is this comment also stale?

ThomasSchellenbergNextCentury (commented):
Looks good to me (tests pass), besides the two places with the stale comments, and leaving a few comments where we want to remove the "override" code in the future.

jaudick (approved):

mcs_history_ingest.py
rartiss55: Thanks for the ticket numbers, I added a comment. And yes, the scenario above is considered valid for the ambiguous case + subtracted balls are not reachable.

tests/test_scorecard.py
```diff
@@ -566,6 +574,77 @@ def test_number_of_times_walked_into_platform_lips(self):
...
+
+    def test_number_of_rewards_achieved_wrong(self):
+        scene_file = mcs_scene_ingest.load_json_file(
+            TEST_FOLDER, TEST_SCENE_NUM_REWARDS)
+        history_file = mcs_scene_ingest.load_json_file(
+            TEST_FOLDER, TEST_HISTORY_NUM_REWARDS_INCORRECT)
+        scorecard = Scorecard(history_file, scene_file)
+        scorecard.calc_num_rewards_achieved()
+        # should be 1, since the ball retrieved was a foil
+        assert scorecard.get_number_of_rewards_achieved() == 0
```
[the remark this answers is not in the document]
rartiss55: The comment is right, this is for a non-ambiguous scene.

tests/test_scorecard.py
rartiss55: Ah mentioned this in another comment, the scene file for this was missing an ambiguous tag/I fixed the test and updated the comment for clarity.

rartiss55:
@ThomasSchellenbergNextCentury I think I addressed your comments.

ThomasSchellenbergNextCentury (approved):

mcs_history_ingest.py
brianpippin: So I think this all looks great but I am curious about this use case and whether we think it is correct.  Say the AI picks up Ball A, and then goes to pick up Ball B, but for some reason doesn't think he can hold both balls, so it drops Ball A, and then picks up Ball B.  This code would mark this as incorrect, but the curious question is whether it should be marked incorrect.  (I honestly don't even know if the API handles the pickup/drop question either, it was just something I was thinking, perhaps it is fine for this evaluation, but just wanted to throw this out there).  Otherwise I think it looks great.

ThomasSchellenbergNextCentury: Hmm... That might be a good question for KDL?

rartiss55: Should be checking unique ids for Pickup/ignoring drops for all multi-reward.

ThomasSchellenbergNextCentury (approved):
Latest changes look good. Tests pass.

PR #118, opened 2023-03-08
MCS-1594 container opened shell game set rotation

jaudick:


scorecard/README.md
```diff
@@ -213,6 +214,68 @@ If they open a wrong one or in the wrong order then the scene ends.
...
+The lanes are labed 1 through 5 and correspond to a global x position.
+* 1 = -1.5 
+* 2 = -0.75
+* 3 = 0
+* 4 = 0.75
+* 5 = 1.5
+
+#### Shell Game Opened Container
+
+For shell game scenes the agent needs to track containers
```
rartiss55, line 268: Thanks for the notes here!

scorecard/scorecard.py
```diff
@@ -1194,3 +1232,182 @@ def calc_imitation_order_containers_are_opened_colors(self):
...
+            ''' 
+            Determine the door the performer opened in the following scenes
+            with the options available:
+            Trajectory - Left, Right
+            InteractiveCollision - Left, Right
+            Solidity - Left, Middle, Right
+            SupportRelations - Left, Middle, Right
+            '''
+            steps_list = self.history['steps']
+            door_opened = ''
```
rartiss55, line 1390: I think it might be good to make door_opened = None so that "N/A" will display in the UI instead of an empty string.

scorecard/scorecard.py
```diff
@@ -1194,3 +1232,182 @@ def calc_imitation_order_containers_are_opened_colors(self):
         self.order_containers_are_opened_colors = \
             order_containers_are_opened_colors
         return self.order_containers_are_opened_colors
+
+    def calc_set_rotation(self):
+        ''' 
+        Determine the container the performer opened in set rotation scenes
```
rartiss55, line 1238: Would it make sense to return (None, None) here if this isn't a set rotation scene?

scorecard/scorecard.py
```diff
@@ -1194,3 +1232,182 @@ def calc_imitation_order_containers_are_opened_colors(self):
...
+            1: (0, 2.62),
+            2: (1.62, 1),
+            3: (0, 0.62),
+            4: (-1.62, 1),
+            5: (0, 1)
+        }
+        set_rotation_opened_container_position_absolute = ''
+        set_rotation_opened_container_position_relative_to_baited = ''
+        self.set_rotation_opened_container_position_absolute = None
+        self.set_rotation_opened_container_position_relative_to_baited = None
```
rartiss55, line 1260: Would it make sense to just return the set rotation variables here as None if this wasn't a set rotation scene?

scorecard/scorecard.py
```diff
@@ -1194,3 +1232,182 @@ def calc_imitation_order_containers_are_opened_colors(self):
...
+
+    def calc_shell_game(self):
+        ''' 
+        Determine the container the performer opened in shell game scenes
+        '''
+        steps_list = self.history['steps']
+        shell_game_baited_container = None
+        shell_game_opened_container = None
+        self.shell_game_baited_container = shell_game_baited_container
+        self.shell_game_opened_container = shell_game_opened_container
```
rartiss55, line 1344: Same question as above: does it make sense to just return the shell game variables here as None if this isn't a shell game scene?

scorecard/scorecard.py
```diff
@@ -1194,3 +1232,182 @@ def calc_imitation_order_containers_are_opened_colors(self):
...
+                                    cl['relative_to_baited']
+                                break
+                if found_container:
+                    break
+            self.set_rotation_opened_container_position_absolute = \
+                set_rotation_opened_container_position_absolute
+            self.set_rotation_opened_container_position_relative_to_baited = \
+                set_rotation_opened_container_position_relative_to_baited
+        except:
+            pass
```
ThomasSchellenbergNextCentury, line 1331: What might raise an exception here?

scorecard/scorecard.py
```diff
@@ -1194,3 +1232,182 @@ def calc_imitation_order_containers_are_opened_colors(self):
...
+                            if resolved_obj_id == cl['id'] or resolved_obj_id == cl['lid']:
+                                shell_game_opened_container = find_shell_game_container_start_end(cl)
+                                found_container = True
+                                break
+                if found_container:
+                    break
+            self.shell_game_baited_container = shell_game_baited_container
+            self.shell_game_opened_container = shell_game_opened_container
+        except:
+            pass
```
ThomasSchellenbergNextCentury, line 1376: Same question as above: What might raise an exception here?

tests/test_scorecard.py
```diff
@@ -52,6 +52,54 @@
 TEST_SCENE_IMITATION = "imitation_eval_5_ex_1.json"
 TEST_HISTORY_IMITATION = "imitation_eval_5_ex_1_history.json"
 
+TEST_SCENE_SET_ROTATION_TURNTABLE_MOVES_270_MIDDLE_BAITED_MIDDLE_PICKED = "set_rotation/setrotation_0001_17_I1_debug.json"
+TEST_HISTORY_SET_ROTATION_TURNTABLE_270_COUNTER_CLOCKWISE = "set_rotation/setrotation_0001_17_history.json"
```
ThomasSchellenbergNextCentury, line 56: I love all of the thorough testing here! Side note: We should probably integrate pre-commit in this repo, with the same style rules we're using in our other python repos. (I noticed that some lines are very long.)

scorecard/scorecard.py
jaudick: Since the commitment to always ingesting debug files was not set when I made this there was not way to check if a scene was a set rotation scene. Now that we always have debug scenes I can use
```
self.scene['goal']['sceneInfo']['tertiaryType'] != "set rotation"
```
and return the values after. So your suggestion would also remove the try except finally as well that surrounds this. I will implement this change.

scorecard/scorecard.py
jaudick: Yes, comment above^

scorecard/scorecard.py
jaudick: Copying my Rachel comment:

> Since the commitment to always ingesting debug files was not set when I made this there was not way to check if a scene was a set rotation scene. Now that we always have debug scenes I can use
> ```
> self.scene['goal']['sceneInfo']['tertiaryType'] != "set rotation"
> ```
> and return the values after.

Since these scenes have debug properties on objects that are specific to this scene, the exception handling would exit the method when trying to get debug properties that didn't exist. The scorecard calls all of the scorecard methods. Scenes that were not set rotation or shell game scenes would hit the exception case. But now that we agreed on debug scenes this can be removed because we can check the scene type.

scorecard/scorecard.py
jaudick: Comment above^

jaudick:
@rartiss55 @ThomasSchellenbergNextCentury I implemented changes and merged master

ThomasSchellenbergNextCentury (approved):

PR #119, opened 2023-03-08
Fix correct_platform_side for scorecard (MCS-1713).

rartiss55:
When I was checking if this was fixed with Thomas' patch for the Python API, realized there was still a small issue here. Sometimes the targetSide tag exists in a scene but is null, so wanted to make sure we aren't calculating it for non-applicable scenes.

### DIFF
--- a/scorecard/scorecard.py
+++ b/scorecard/scorecard.py
@@ -922,7 +922,7 @@ class Scorecard:
         # Does this scene have a targetSide? If not, return
         # correct_platform_side (currently set to None).
         goal = self.scene.get('goal')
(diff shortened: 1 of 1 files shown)


ThomasSchellenbergNextCentury (approved):
I was confused at first because I didn't think there would be any null target sides, but then I saw your other PR: https://github.com/NextCenturyCorporation/mcs-private/pull/387

rartiss55:
Updated per comment on the linked mcs-private PR.

ThomasSchellenbergNextCentury (approved):

PR #122, opened 2023-03-22
Fix for set rotation scorecard metric.

ThomasSchellenbergNextCentury:


scorecard/scorecard.py
```diff
@@ -1434,17 +1434,17 @@ def calc_set_rotation(self):
...
                                 break
                 if found_container:
                     break
             self.set_rotation_opened_container_position_absolute = \
-                set_rotation_opened_container_position_absolute
+                self.set_rotation_opened_container_position_absolute
             self.set_rotation_opened_container_position_relative_to_baited = \
-                set_rotation_opened_container_position_relative_to_baited
+                self.set_rotation_opened_container_position_relative_to_baited
         except:
```
rartiss55, line 1448: So I removed this except block here and got this error locally:
  File "/Users/rachel.artiss_cn/MCS/mcs-ingest/scorecard/scorecard.py", line 1391, in calc_set_rotation
    rotation_direction = self.scene['goal']['sceneInfo']['rotation']
KeyError: 'rotation'

Is this another instance of something that should be in the scene file but isn't?

ThomasSchellenbergNextCentury: Should be fixed now. You can run the new `run_scorecard.py` script to test it yourself! :)

rartiss55 (approved):
I don't think the self.set_rotation property update is technically needed now that there's a check for rotation key, but this should work!

jaudick (approved):

brianpippin (approved):

phile-caci (approved):

PR #124, opened 2023-04-13
MCS-1757: Script to add domain type to scenes for Evaluation Status page over haul

brianpippin:


phile-caci (commented):
Hey @brianpippin ,
<del>I'm seeing this in unit test:</del>(ran with unittest, but this passes in pytest)
```
.............................Not computing dist, tilt is 0.0
.No target by step data for scene occluders_0001_17
........................No target by step data for scene occluders_0001_17
..................
----------------------------------------------------------------------
```

I also tested a local ingest of the validation debug scenes.  That went into ok.
I then ran the local ingest of OPICS v2 validation history files and saw this error:

```
python local_history_ingest.py --folder /home/ple/Desktop/eval-6-val-history
/home/ple/mcs/mcs-ingest/venv/lib/python3.8/site-packages/shapely/measurement.py:74: RuntimeWarning: invalid value encountered in distance
  return lib.distance(a, b, **kwargs)
```

My ingest knowledge is still very narrow, so please let me know if I missed something, thanks!

ThomasSchellenbergNextCentury (approved):
I haven't tried running this script myself, but the code looks fine. Also, my tests pass without warnings.

phile-caci:
Looks like my shapely was on 2.0.1, reverting to 1.8.5 worked.  I'll update the requirements to pin the shapely version.

phile-caci (approved):
Retested after pinning shapely, looks good.

PR #130, opened 2023-04-26
MCS-1784: Fix agents imitation incorrect scoring

brianpippin:


scripts/_0_6_5_6_fix_agent_scoring_wrong_ground_truth.py
```diff
@@ -0,0 +1,269 @@
...
+    print("Begin looping through scene file names.")
+    for history_file_name in scene_file_names:
+        history_files = results_collection.find({"category_type": "agents imitation", "name": history_file_name})
+
+        # Should be 3 history files for each name, one for each performer
+        for history_record in history_files:
+            history_record["score"]["ground_truth"] = 1 if history_record["score"]["ground_truth"] == 0 else 0
+
+            paired_record = mcs_history_ingest.return_agency_paired_history_item(
+                        client, db_string, history_record)
```
ThomasSchellenbergNextCentury, line 254: @brianpippin Don't we need to make sure the `ground_truth` for the paired record has already been switched before rescoring them?

ThomasSchellenbergNextCentury: Or will each pair just get rescored twice: first incorrectly, then correctly?

brianpippin: Each pair will get rescored twice, the first time will be incorrect, the second time will be correct since both ground_truths will have been flipped at that point.  Better to just have it run through this way then make assumptions.  All this code is copied from code we have in history ingest and used in the other pair rescoring.

ThomasSchellenbergNextCentury: Sounds good -- Just wanted to clarify. Thank you.

ThomasSchellenbergNextCentury (approved):

brianpippin:
I ran this locally and all of the performers had a considerable improvement in scoring as well :)

PR #132, opened 2023-05-24
MCS-1804 - Add first step information for performer + target start positions to csv

rartiss55:
If approved, will also have to run update_collection_keys_if_missing.py for this + regenerate the results csv. 

What I ended up doing was just omitting most "steps." keys, since mongoexport isn't saving that array anyway, and am explicitly saving the first step for the fields specified in the ticket like so:
    'steps.0.output.position.x',
    'steps.0.output.position.y',
    'steps.0.output.position.z',
    'steps.0.output.rotation',
    'steps.0.output.target.position.x',
    'steps.0.output.target.position.y',
    'steps.0.output.target.position.z',

Let me know if that makes sense.

### DIFF
--- a/create_collection_keys.py
+++ b/create_collection_keys.py
@@ -1,5 +1,17 @@
 import re
 
+# For the following step array fields, we want to save the 
+# value contained in the first step.
+STEPS_OUTPUT_FIELDS_TO_SAVE = [
+    "steps.output.position.x",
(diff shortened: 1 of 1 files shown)


rartiss55:
In case anyone's curious, realized there's an issue doing it this way + that I should probably handle it on the UI side.

PR #133, opened 2023-05-30
Mcs 1827 exclude opics imitation

phile-caci:
Excludes scenes from UI score where OPICS retrieves the soccer ball from the ceiling.  

Set 'score.weighted_score' and 'score.weighted_score_worth' to 0 for  performer = OPICS and category_type = "imitation task" when scene files ("name") was one of the affected scenes. Reviewed the rest of the scores against prod and looks like nothing else was affected.

Rebuilt my local UI with just Eval 6 (mcs-ui) and started local containers, then in mcs-ingest run `python deployment_script.py`

Before:
image

After:
image

### DIFF
--- a/deployment_script.py
+++ b/deployment_script.py
@@ -12,7 +12,8 @@ def check_version(mongoDB):
     collection = mongoDB[VERSION_COLLECTION]
     version_obj = collection.find_one()
     # return true if it is a newer db version
-    return db_version > version_obj['version']
+    return True
+    # return db_version > version_obj['version']
 
 
 def update_db_version(mongoDB):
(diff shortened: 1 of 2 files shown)


ThomasSchellenbergNextCentury (approved):
The code LGTM.

Will the `"Current scene is not included in passing score"` message appear for these? Does it automatically appear if the score worth is zero?

rartiss55:
> The code LGTM.
> 
> Will the `"Current scene is not included in passing score"` message appear for these? Does it automatically appear if the score worth is zero?

Tested locally -- looks like its appearing for these scenes + they're excluded from the Test Overview page.

PR #134, opened 2023-06-15
MCS-1787 - Remove special handling for multi-retrieval scenes

rartiss55:
Shouldn't be necessary anymore with recent updates to the Python API + reward is now calculated correctly for multi-retrieval.

There was a note in the ticket about backwards compatibility, I opted to just take the multi-retrieval override stuff out because it added complexity to the code that I don't we really need with one eval remaining, as well as the test cases (having to track different scene/history files).


mcs_history_ingest.py
```diff
@@ -522,25 +502,10 @@ def build_new_step_obj(
...
-                    multi_retrieval_rewards_picked_up.append(resolved_obj)
-
-            if(total_targets_multi_retrieval == len(multi_retrieval_rewards_picked_up) and interactive_goal_achieved != 1):
-                interactive_goal_achieved = 1
-
         if "target_visible" in step:
             new_step["target_visible"] = step["target_visible"]
 
-        target_keys = ['target', 'target_1', 'target_2']
+        target_keys = ['targets', 'target', 'target_1', 'target_2']
```
rartiss55, line 508: We aren't saving the target array to the mongo database, added this to the array to do just that.

ThomasSchellenbergNextCentury (commented):
Do we need to save any of this code for backwards compatibility? Would we ever need to reingest the Eval 6 multi-retrieval scenes?

ThomasSchellenbergNextCentury (approved):
Sounds like we don't care about backwards compatibility (that's fine with me).

PR #137, opened 2023-07-26
MCS-1912 - correct scorecard variables for trajectory and interactive collisions

rartiss55:
Koleen initially mentioned correct_platform_side as being incorrect, realized that was also the case for correct_door_opened (all of the latter were set to false). 

Ran this locally, things looked correct for the results I checked.

### DIFF
--- a/scorecard/scorecard.py
+++ b/scorecard/scorecard.py
@@ -981,7 +981,9 @@ class Scorecard:
             target_side = goal['sceneInfo']['toolChoiceValidSide']
         elif (
             'sceneInfo' in goal and 'relation' in goal['sceneInfo'] and
-            goal['sceneInfo']['relation'] in ['sameSide', 'oppositeSide']
+            goal['sceneInfo']['relation'] in ['sameSide', 'oppositeSide'] and
+            'type' in goal['sceneInfo'] and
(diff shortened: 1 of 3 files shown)


scorecard/scorecard.py
```diff
@@ -976,6 +979,39 @@ def calc_correct_platform_side(self):
...
+            # if for whatever reason, we can't find the
+            # throwing device, return
+            if len(throwing_device) == 0:
+                return self.correct_platform_side
+
+            relation = goal['sceneInfo']['relation']
+            collision = goal['sceneInfo']['type']
+            x_pos = throwing_device[0]['shows'][0]['position']['x']
+
+            is_target_same_side = relation == 'sameSide' and collision == 'noCollision'
```
rartiss55, line 1002: Mentioned at standup, but I believe checking the position of the thrower (as opposed to using the initial target position) against the "opposite/sameside" tag and whether or not there's a collision still works with how its set up here (scenes tagged "collision" should end up with the target on the opposite side of the thrower). Please let me know though if there's something I missed.

ThomasSchellenbergNextCentury: Yeah, that makes sense. I forgot that "sameSide" was short for "target and shooter/thrower on the same side".

PR #138, opened 2023-08-31
MCS-1935: Update tool use section with new tertiary types 

rartiss55:
See PR/PR comment here: https://github.com/NextCenturyCorporation/mcs-private/pull/503#pullrequestreview-1603404681

Note that this section will need to be updated again prior to Eval 7, but that will be covered by MCS-1752.


rartiss55:
> This doesn't need to be backwards compatible, right?

@ThomasSchellenbergNextCentury No, we usually just update it every eval / shouldn't have to worry about anything since the Eval 6 rerun is complete.

PR #139, opened 2023-10-04
MCS-1956 - Adding distance calculation between start performer/target positions to history record

rartiss55:
Tested with "Eval 7" labeled data to make sure it appears in the CSV, which reminded me that when we run Eval 7, we'll have to run the  update_collection_keys_if_missing.py script after we ingest a few files. 

UI PR here: https://github.com/NextCenturyCorporation/mcs-ui/pull/149


phile-caci (approved):
Looks good, all test pass, Setup and locally and ingested a single file to test.  New field is in the csv.

image

PR #140, opened 2023-10-18
MCS-1971 - Add new scorecard values for Eval 7

rartiss55:
New scorecard properties:
- For multi tool: how many tools were used
- For multi tool: was the hooked tool rotated or not
- For multi tool: was the straight tool rotated or not
- For any lava scene: whether or not performer stepped on lava

For anyone testing out the data here in ingest + the UI, note that any multi tool specific scenes I marked as "eval_7", but the lima and mars scenes were taken from eval_6 data.

See UI PR here: https://github.com/NextCenturyCorporation/mcs-ui/pull/151


tests/test_data/test_eval_3-5_level2_baseline_juliett_0001_01.json
rartiss55, line 1: Only updated for the unit test (since steps_on_lava was added post-Eval 3.5).

scorecard/scorecard.py
```diff
@@ -946,9 +959,27 @@ def calc_tool_usage(self):
                 resolved_obj = get_relevant_object(output)
                 if resolved_obj.startswith('tool') and return_status == 'SUCCESSFUL':
                     tool_usage[action] += 1
+
+                    if is_multi_tool:
```
ThomasSchellenbergNextCentury, line 963: Is this condition necessary? Might these metrics be useful for normal Tool Use scenes as well?

scorecard/scorecard.py
```diff
@@ -946,9 +959,27 @@ def calc_tool_usage(self):
...
+                        unique_tools.add(resolved_obj)
+
+                        if action in ['RotateObject', 'TorqueObject']:
+                            tool_object = [obj for obj in self.scene['objects']
+                                if obj['id'] == resolved_obj]
+                            if(len(tool_object) > 0):
+                                if(is_straight_rotated == False and tool_object[0]['type'].startswith('tool_rect')):
+                                    is_straight_rotated = True
+                                if(is_hooked_rotated == False and tool_object[0]['type'].startswith('tool_hooked')):
```
ThomasSchellenbergNextCentury, line 972: This needs to include `tool_isosceles`

rartiss55: Added `tool_isosceles` to hooked case.

scorecard/scorecard.py
rartiss55: I personally thought it would be redundant (there's only one accessible tool in other scenes I thought so the other metrics should cover it). Let me know if I'm missing something though.

phile-caci (approved):
Looks good, all test pass.

Tool Usage (Multi):
image

Step on Lava:
image

scorecard/scorecard.py
ThomasSchellenbergNextCentury: You're right, I thought this statement encapsulated all of the tool usage metrics, but it doesn't. My mistake.

PR #141, opened 2023-10-25
MCS-1970: Update shell game scorecard value to match more with set rotation

rartiss55:
Changing "shell_game_baited_container" to "shell_game_opened_container_position_relative_to_baited"


scorecard/scorecard.py
```diff
@@ -1583,10 +1584,19 @@ def calc_shell_game(self):
                 'moves': obj.get('moves')
             }
             for obj in self.scene['objects'] if obj['type'] == 'separate_container']
-        for cl in containers_and_lids:
-            if cl['isTargetContainer']:
-                shell_game_baited_container = find_shell_game_container_start_end(cl)
-                break
+
+        if('baitedContainerMovement' in self.scene['goal']['sceneInfo']):
```
rartiss55, line 1588: I added some backwards compatibility here only because this change wouldn't have made it into the validation data set.

scorecard/scorecard.py
```diff
@@ -1597,13 +1607,44 @@ def calc_shell_game(self):
                     for cl in containers_and_lids:
                         if resolved_obj_id == cl['id'] or resolved_obj_id == cl['lid']:
                             shell_game_opened_container = find_shell_game_container_start_end(cl)
+                            opened_ctr_end_pos = shell_game_opened_container[-1]
```
rartiss55, line 1610: Tried to line this up somewhat with what we have for set rotation, but let me know if something doesn't sound quite right:

If the baited container was opened, the relative keyword will be "baited"

For all other cases:
For two container scenes when non-baited is opened: "left" or "right"
For three container scenes when non-baited is opened:
- if baited is on an end: "middle" or "opposite"
- if baited is in the middle: "left" or "right"

phile-caci (approved):
Looks good, all test pass, ran a shell game and ingested scene and history:

image

rartiss55:
> Looks good, all test pass, ran a shell game and ingested scene and history:
> 
> image

@phile-caci These are the old values (the first one in that picture should be a keyword instead). Could you double check the branch?

phile-caci:
> > Looks good, all test pass, ran a shell game and ingested scene and history:
> > image
> 
> @phile-caci These are the old values (the first one in that picture should be a keyword instead). Could you double check the branch?

ah k.. checking

phile-caci:
@rartiss55 , sry, here's the correct screenshot:
image

ThomasSchellenbergNextCentury (approved):
The code looks good to me.

PR #142, opened 2023-11-07
MCS-1985 - Updating set rotation scorecard items for 5 container case

rartiss55:
Incorporating 5 container case introduced for Eval 7. 

In case you're worried about the files changed amount, most of those are test files.


scorecard/scorecard.py
```diff
@@ -1458,20 +1458,30 @@ def calc_set_rotation(self):
...
+                |
+                3
+            Performer
+
         """
         absolute_positions = {
             1: (0, 2.62),
             2: (1.62, 1),
-            3: (0, 0.62),
+            3: (0, -0.62),
```
rartiss55, line 1478: Also updated this value for 3 -- realized it should be negative.

scorecard/scorecard.py
```diff
@@ -1513,22 +1523,117 @@ def calc_set_rotation(self):
...
+                        right_x = absolute_positions.get(2)[0]
+
+                        # need to reverse +/- of other containers in this case
+                        flipped = (
+                            (rotation_direction.startswith(('counter')) and rotation == 270) or
+                            (rotation_direction.startswith(('clock')) and rotation == 90) or
+                            rotation == 180)
+
+                        """
+                            Breakdown of labeling for five container case:
```
rartiss55, line 1554: Here's the labeling breakdown -- the +/- depends on if the baited container is on the left/nearest vs on the right/furthest away post-rotation.

tests/test_scorecard.py
```diff
@@ -89,6 +89,48 @@
 TEST_SCENE_SET_ROTATION_1_TABLE_90_CL_LEFT_BAITED_FAR_PICKED = "set_rotation/setrotation_0003_01_A1_debug.json"
 TEST_HISTORY_SET_ROTATION_1_TABLE_90_CL_LEFT_BAITED_FAR_PICKED = "set_rotation/setrotation_0003_01_history.json"
 
+TEST_SCENE_SET_ROTATION_5_TEST_5_TABLE_MOVES_90_CCW = "set_rotation/set_rotation_0005_33_M1_debug.json"
```
rartiss55, line 92: Added quite a few test files -- I don't have every rotation represented but wanted to make sure the labeling was right for many of the baited positions vs opened container cases since its so variable.

phile-caci (approved):
LGTM, all the test pass.

Ran and ingested a scene with 5 containers. The baited container went from 9 to 6 (90 degrees) and then opened the baited container:

image

scorecard/README.md
```diff
@@ -33,6 +33,8 @@ The actions that are counted (as of Eval 5):
 * Determine what order the agent opened the containers in an imitation task.
 Order is determined by color (green, blue, red, etc.)
 * Determine what door the agent opened (left, middle, right)
+* Determine (for multi tool use) how many unique tools were touched and which tools were rotated.
```
rartiss55, line 36: Realized I completely forgot about this -- added all the eval 7-related info.

ThomasSchellenbergNextCentury: Thanks!

tests/test_scorecard.py
ThomasSchellenbergNextCentury: Seems good to me!

ThomasSchellenbergNextCentury (approved):
All the tests pass, and the code changes look sensible.

PR #143, opened 2023-11-17
MCS - 1752 - Updated passing cells

NDCACI:


mcs_history_ingest.py
```diff
@@ -78,7 +78,13 @@
                  "G2", "I1", "I2", "K1", "K2"],
     "gravity support": ["A1", "B1", "C1", "D1", "I1", "J1", "K1", "L1",
                         "M1", "N1", "O1", "P1", "S1", "T1", "W1", "X1",
-                        "Y1", "Z1", "AA1", "CC1", "SS1", "TT1"]
+                        "Y1", "Z1", "AA1", "CC1", "SS1", "TT1"],
+    "set_rotation":["C1", "F1", "I1", "L1", "C2", "F2", "I2", "L2", "M1", "N1",
```
rartiss55, line 82: The tertiaryType field doesn't have underscores, so these categories would actually be "set rotation" and "multi tool use" respectively.

rartiss55 (dismissed):
Left comment about underscores.

mcs_history_ingest.py
```diff
@@ -94,6 +102,7 @@
     "shell game": ["A1", "B1", "E1", "F1", "I1", "J1"],
     "set rotation": ["B1", "C1", "E1", "F1", "H1", "I1", "K1", "L1", "B2",
```
rartiss55, line 103: I'm guessing this file got really stale and was missing the existing "set rotation"? Now there are two (lines 88 and 103), and it looks like the multi tool use line needs a comma after it.

rartiss55 (approved):

ThomasSchellenbergNextCentury (approved):
LGTM.

Agent Identification is irrelevant since we don't generate C or G scenes anyway but it won't hurt us to keep the changes.

